Issue #12, opened 2016-06-08
Global var and function


zaglex:
In expirationd.lua there are suspicious global function: 'expirationd_run_task_obsolete', and global variable: 'retval' in expirationd_task_stats.  I suppose they should be made local.

kostja:
Hi, thanks for the issue! Care to submit a pull request?

bigbes:
Merged! Thanks.

PR #13, opened 2016-06-13
Fix global scope issue (#12)

zaglex:
Scope of next items changed to local:
1. Function expirationd_run_task_obsolete
2. Var 'retval' in expirationd_task_stats


bigbes:
Thank you for your PR, I've merged it by hand with some additions.

PR #40, opened 2020-05-18
Iterate over the space without loading all the data into lua memory

derElektrobesen:
After tarantool with fix for issue https://github.com/tarantool/tarantool/issues/4556 upgrading, our tarantool has crashed with following lines in log file:
```
2020-05-06 20:42:37.243 [13299] main/117/worker of "exp_toks" C> not enough memory
2020-05-06 20:42:37.243 [13299] main/117/worker of "exp_toks" C> #2 tree_index_iter (upvalue), ./expirationd.lua:96
2020-05-06 20:42:37.243 [13299] main/117/worker of "exp_toks" C> #3 do_worker_iteration (field), ./expirationd.lua:124
2020-05-06 20:42:37.243 [13299] main/117/worker of "exp_toks" C> #4 (null) (), ./expirationd.lua:163
```

Expirationd working with a space where possibly could be a large number of large tuples. So loading a large number of this tuples into lua memory is inefficient and could be a reason of tarantool crashes.

### DIFF
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -83,20 +83,15 @@ local function suspend(scan_space, task)
 end
 
 local function tree_index_iter(scan_space, task)
-    -- iteration with GT iterator
-    local params = {iterator = 'GT', limit = task.tuples_per_iteration}
-    local last_id
-    local tuples = scan_space.index[0]:select({}, params)
-    while #tuples > 0 do
-        last_id = tuples[#tuples]
-        for _, tuple in ipairs(tuples) do
-            expiration_process(task, tuple)
-        end
-        local key = construct_key(scan_space.id, last_id)
-        tuples = scan_space.index[0]:select(key, params)
-        suspend(scan_space, task)
-    end
+    local n_processed = 0
+    for _, tuple in scan_space.index[0]:pairs({}, {iterator = 'GT'}) do
+        expiration_process(task, tuple)
 
+        n_processed = n_processed + 1
+        if n_processed % task.tuples_per_iteration == 0 then
+            suspend(scan_space, task)
+        end
+    end
 end
 
 local function hash_index_iter(scan_space, task)


Totktonada:
Related: https://github.com/tarantool/tarantool/commit/2160e4cbdb90caff0a3c766bd92003d1f7d67d7b

Totktonada:
I don't sure that this limited select is the root cause of the `not enough memory` error. I asked Pavel and he said that `tuples_per_iteration` parameter is default (1024).

(For record: default `full_scan_time` (3600), `is_tuple_expired` like `function(t) t[EXPIRE_TIME_FIELD] + CONSTANT < fiber.time() end`, default `process_expired_tuple` (which deletes a tuples from a space).)

Size of tuples should not affect lua memory, because only pointer is counted in lua memory (see https://github.com/tarantool/tarantool/issues/1207). Yep, when a field is accessed, it is pushed to lua memory, but those values are likely temporary and may be collected soon. They should be usually not large: primary key fields and `EXPIRE_TIME_FIELD` are accessed in the Pavel's scenario.

So, again, I doubt that this select is problematic per se.

Let's consider PR as optimization now. First, it is based on stable tree index iterators, that was introduced in [1.7.4-420-g2160e4cbd](https://github.com/tarantool/tarantool/commit/2160e4cbdb90caff0a3c766bd92003d1f7d67d7b) and so the new code may give unpredictable results (missed tuples, double visited tuples) on lower versions. Okay, we can say that it is okay for us or wrap it with a tarantool version check. What we'll give from the new way to iterate over tuples on newer tarantool versions:

* We'll save Lua memory for `tuples_per_iteration` * `cdata size`, which looks tiny if `tuples_per_iteration` value is default or somewhat adequate.
* Tuples that are inserted in a middle of iteration after currently processed one will be visited within this iteration. This behaviour change looks not important for typical expirationd usage (visit all tuples within specified time period).

Maybe @alyapunov would correct me.

Totktonada:
To be honest, I don't want to break compatibility with old tarantool versions and I don't see any prove that the pointed code may be problematic (not just was hit by OOM, because some other code eats all available Lua memory). I would accept the change as the optimization if it would be guarded with a tarantool version check ([example](https://github.com/tarantool/tuple-keydef/blob/10c00a9aa289bb16446239edfa056ee48b467161/test/tuple_keydef.test.lua#L11-L40)).

NB: The obvious compatible change to save a bit Lua memory would be:

```diff
diff --git a/expirationd.lua b/expirationd.lua
index fc71068..1d28daf 100644
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -92,9 +92,11 @@ local function tree_index_iter(scan_space, task)
         for _, tuple in ipairs(tuples) do
             expiration_process(task, tuple)
         end
+        tuples = nil
         local key = construct_key(scan_space.id, last_id)
-        tuples = scan_space.index[0]:select(key, params)
+        last_id = nil
         suspend(scan_space, task)
+        tuples = scan_space.index[0]:select(key, params)
     end
 
 end
```

Or, a bit more readable:

```diff
diff --git a/expirationd.lua b/expirationd.lua
index fc71068..ce3a64c 100644
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -85,16 +85,15 @@ end
 local function tree_index_iter(scan_space, task)
     -- iteration with GT iterator
     local params = {iterator = 'GT', limit = task.tuples_per_iteration}
-    local last_id
     local tuples = scan_space.index[0]:select({}, params)
     while #tuples > 0 do
-        last_id = tuples[#tuples]
         for _, tuple in ipairs(tuples) do
             expiration_process(task, tuple)
         end
-        local key = construct_key(scan_space.id, last_id)
-        tuples = scan_space.index[0]:select(key, params)
+        local next_key = construct_key(scan_space.id, tuples[#tuples])
+        tuples = nil
         suspend(scan_space, task)
+        tuples = scan_space.index[0]:select(next_key, params)
     end
 
 end
```

But I'm not fan of changing code just to fit my personal taste. There should be a good reason. Otherwise sooner or later we'll meet some subtle behaviour change that hits a user in a non-obvious way.

I'll close the PR if it will not receive any updates. Sorry.

Totktonada:
We had a short discussion with @olegrok around this change. I want to share some points from it.

* We cannot push the change as a fix of overwhelming memory consumption issue (see above).
* It does not looks practical to push the change as optimization of memory consumption (see above).
* However it may have some value as refactoring. There are several points that I would highlight in this context.

How the implementation looks now:

* A hash index: open an iterator for a complete space traversal, sleep during the traversal.
* A tree index: fetch tuples in batches using GT iterator and limit, sleep between batches processing.
* (rtree / bitset are not supported, but there is no corresponding error message)
* A vinyl index iterator (only tree is supported by vinyl): looks the same as for memtx tree index (but customized for working with unknown / imprecise space length).

If we can unify this logic for all index types and engines, it would greatly improve the code and it would be worthful refactoring (even if we'll break compatibility with tarantool 1.7.4 and below). However we should understand an index iterator guarantees first: I filed https://github.com/tarantool/doc/issues/2102 regarding this. At least I would want to understand whether a vinyl iterator is stable in the same sense as a memtx tree iterator after 1.7.5 (and why the expirationd code does not lean on it if it is so).

What else we can or should consider if we'll do a refactoring of this kind:

* Terms in the code. The term 'iteration' is used for a complete space traversal, however in the API of the module it relates to a batch. A space traversal is referred by `.*full_scan.*` options / functions in the API, `*iteration*` refers to a batch. It would be good to stop using the same term for different things internally, that's confusing.
* Consider upcoming feature #50. The aspects that are relevant here are the following:
  * ability to pass a key to start a space traversal,
  * ability to choose ascending or descending order,
  * ability to traverse using a secondary index.

PR #46, opened 2021-02-17
Remove Ubuntu Eoan from CI

LeonidVas:
Ubuntu Eoan is EOL and fails in CI on apt repositories updating with 'does not have a Release file' errors.


Totktonada (approved):
LGTM.

PR #49, opened 2021-04-13
test: check hotreload

### LINKED ISSUE #54: Add test for reloading the module with external iteration state via Cartridge hotreload
Use case:

A user creates an expiration task, for which the expiration check callback or the iteration function saves some state into the globals or `cartridge.vars` (e.g. iterator state or the last tuple value, etc). In case of reloading the module in a Cartridge role, the user must have the ability to preserve the state for continuing the iteration with it after the reload.

AC:

New integration test added for the above scenario and is passing in the CI

ArtDu:
The class of problems that this test catches is the expirationd integration with cartridge. When hot-reloading roles (that is, when call require("cartridge.roles").reload() ), expirationd should exit gracefully. All fibers should terminate and the task list should be cleared.

These tests will also be used as an example of using hot reload and expirationd

tests:
- check how expiraiond will behave when reloading a role without writing additional code in role stop
- transactions will be processed correctly if we specify the use of batch transactions in expirationd.start
- reloading the role with saving the iteration state and starting the expirationd task from the moment the last role ended

Closes #54

### DIFF
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -272,6 +272,7 @@ local function create_task(name)
         process_while                  = constants.process_while,
         iterator_type                  = constants.iterator_type,
         atomic_iteration               = constants.atomic_iteration,
+        force                          = constants.force
     }, { __index = Task_methods })
     return task
 end
--- a/test/integration/hotreload_test.lua
+++ b/test/integration/hotreload_test.lua
@@ -1,0 +1,304 @@
+local utils = require("cartridge.utils")
+local fio = require("fio")
+local t = require("luatest")
+local g = t.group("hot_reload")
+
+local helpers = require("test.helper")
+
+local function reload_myrole(fn)
+    -- For the sake of string.dump() function must have no upvalues.
+    -- https://www.lua.org/manual/5.1/manual.html#pdf-string.dump
+    utils.assert_upvalues(fn, {})
+
+    local ok, err = g.srv.net_box:eval([[
+        package.preload["mymodule"] = loadstring(...)
+        return require("cartridge.roles").reload()
+    ]], {string.dump(fn)})
+
+    t.assert_equals({ok, err}, {true, nil})
+end
+
+g.after_each(function()
+    g.srv.net_box:eval("box.space.origin:truncate()")
+end)
+
+local function get_fiber_names()
+    local fiber_info = g.srv.net_box:eval("return require('fiber').info()")
+    local fiber_names = {}
+    for _, f in pairs(fiber_info) do
+        fiber_names[f.name] = true
+    end
+    return fiber_names
+end
+
+g.before_all(function()
+    local tempdir = fio.tempdir()
+    g.cluster = helpers.Cluster:new({
+        datadir = tempdir,
+        server_command = helpers.entrypoint("srv_reload"),
+        cookie = "secret-cluster-cookie",
+        replicasets = {{
+                           alias = "A",
+                           roles = {"myrole"},
+                           servers = 1,
+                       }},
+    })
+    g.srv = g.cluster:server("A-1")
+    g.cluster:start()
+end)
+
+g.after_all(function()
+    g.cluster:stop()
+    fio.rmtree(g.cluster.datadir)
+end)
+
+function g.test_graceful_shutdown()
+    -- In this test, we will check how the expirationd task
+    -- will behave during cartridge reload, if no additional logic is written.
+    -- It is expected that the task will be completed gracefully
+    -- and will not delete any more data. All fibers in the task will be removed.
+    -- And we will be able to launch new expirationd tasks with role reload
+    -- and they will also work correctly (deleting data)
+
+    reload_myrole(function()
+        return {
+            role_name = "myrole",
+            init = function()
+                local expirationd = require("expirationd")
+                local helpers = require("test.helper")
+
+                box.space.origin:insert({1})
+                expirationd.start("first_task", box.space.origin.id, helpers.is_expired_debug,  {
+                    force = true
+                })
+                return true, nil
+            end,
+            stop = function() end
+        }
+    end)
+    -- check that the task is running
+    t.assert_equals(
+            g.srv.net_box:eval([[
+                local expirationd = require("expirationd")
+                return expirationd.tasks()
+            ]]),
+            {"first_task"}
+    )
+    -- fibers of task
+    local fiber_names = get_fiber_names()
+    t.assert(fiber_names['guardian of "first_task"'])
+    t.assert(fiber_names['worker of "first_task"'])
+    -- we see that the tuple is expired
+    t.assert_equals(
+            g.srv.net_box:eval("return box.space.origin:select()"),
+            {}
+    )
+    -- expirationd task checked the tuple
+    t.assert_equals(
+            g.srv.net_box:eval([[
+                local helpers = require("test.helper")
+                return helpers.iteration_result
+            ]]),
+            {{1}}
+    )
+    reload_myrole(function()
+        return {
+            role_name = "myrole",
+            init = function()
+                local expirationd = require("expirationd")
+                local helpers = require("test.helper")
+
+                -- after reload we will see that tuple {2} has not expired
+                box.space.origin:insert({2})
+
+                expirationd.start("second_task", box.space.origin.id, helpers.is_expired_debug, {
+                    force = true,
+                    process_expired_tuple = function()  end -- don't delete tuple
+                })
+                return true, nil
+            end,
+            stop = function() end
+        }
+    end)
+
+    -- check that the first_task is gone and second_task is running
+    t.assert_equals(
+            g.srv.net_box:eval([[
+                local expirationd = require("expirationd")
+                return expirationd.tasks()
+            ]]),
+            {"second_task"}
+    )
+    fiber_names = get_fiber_names()
+    -- no fibers of first_task
+    t.assert_not(fiber_names['guardian of "first_task"'])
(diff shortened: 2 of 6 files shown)


expirationd.lua
```diff
@@ -502,10 +509,7 @@ local function expirationd_task_stats(name)
     return retval
 end
 
--- kill task
-local function expirationd_kill_task(name)
-    return get_task(name):kill()
-end
+
```
akudiyar, line 512: Do not get this change

test/entrypoint/srv_basic.lua
```diff
@@ -0,0 +1,46 @@
...
+        validate_config = function()
+            return true
+        end,
+        init = function(opts) end,
+        apply_config = function(_, opts) end,
+        stop = function() end,
+    }
+end
+
+
```
akudiyar, line 26: Extra newlines

akudiyar (changes requested):
The changes must be validated with @Totktonada

expirationd.lua
ArtDu: I removed duplicate code from
https://github.com/tarantool/expirationd/blob/6c7ff791ce77a176540ca7a7233e889788bbbd72/expirationd.lua#L485-L487

.github/workflows/test_on_push.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  run-tests-ce:
+    if: |
+      github.event_name == 'push' ||
+      github.event_name == 'pull_request' && github.event.pull_request.head.repo.owner.login != 'tarantool'
```
Totktonada, line 11: I have the better snippet, which should handle PR to a fork gracefully:

https://github.com/tarantool/smtp/blob/322254c83169905bc43cb8a25972fa19a03f72ea/.github/workflows/testing.yml#L12-L17

.github/workflows/test_on_push.yaml
```diff
@@ -0,0 +1,39 @@
...
+      matrix:
+        tarantool-version: ["1.10", "2.2", "2.3", "2.4", "2.5", "2.6", "2.7"]
+      fail-fast: false
+    runs-on: [ubuntu-latest]
+    steps:
+      - uses: actions/checkout@master
+
+      - name: Install requirements for community
+        run: |
+          curl -L https://tarantool.io/installer.sh | sudo VER=${{ matrix.tarantool-version }} bash
```
Totktonada, line 22: If we only need to run tests on Linux on GitHub owned runners on latest tarantool releases within a release series at the moment, it is better to use https://github.com/tarantool/setup-tarantool: it caches dependencies to speed up CI.

.github/workflows/test_on_push.yaml
```diff
@@ -0,0 +1,39 @@
...
+          sudo apt install -y tarantool-dev
+          tarantool --version
+          ./deps.sh
```
Totktonada, line 25: I would not extract it from the CI script: it is easier to read everything in one place.

.github/workflows/test_on_push.yaml
Totktonada: If you aims to implement more powerful CI, I suggest to look at tarantool/smtp: https://github.com/tarantool/smtp/tree/master/.github/workflows

Still not powerful enough, though:

https://github.com/tarantool/smtp/issues/34
https://github.com/tarantool/smtp/issues/35
https://github.com/tarantool/smtp/issues/36

.github/workflows/test_on_push.yaml
```diff
@@ -0,0 +1,39 @@
+name: Run tests
```
Totktonada, line 1: General thoughs (not a review in fact, I just briefly glanced at Ci changes. I plan to do thorough review on the next week):

1. Please, extract CI enhancements into its own PR. It'll simplify our communication around changes. If you need them to ease work on the code: okay, just base your code / testing changes on your CI changes. We'll push CI changes first, rebase code / testing changes upward and push them.
2. I see that you removed RPM / Deb packages building and deployment. If you don't plan to move it to GitHub Actions, please, keep it on Travis CI (with the idea to move in a future).

ArtDu: This PR is small, so it seems to me you can leave this PR with the addition of ci for luatest and one test from it.
I left building and deployment in travis. And new Luatests are launched on gh ci.

.github/workflows/testing.yaml
```diff
@@ -0,0 +1,248 @@
...
+          #
+          # Verify that system's libcurl headers are not required.
+          - tarantool: live/2.8
+            dont_install_system_libcurl_header: true
+          # Applicable for: 2.6.3, 2.7.2+, 2.8.1+, 2.9+ (gh-4559).
+          #
+          # Verify that tarantool's libcurl headers are preferred
+          # when the system provides them too.
+          - tarantool: live/2.9
+            break_system_libcurl_header: true
```
Totktonada, line 52: Do you ever read the code you copied? If so, explain how libcurl is relevant here. If not, please, never do attempt to proceed with a patch this way again.

.github/workflows/testing.yaml
```diff
@@ -0,0 +1,248 @@
...
+#          - 2.5.1
+#          - 2.5.2
+#          - 2.5.3
+#          - 2.6.1
+#          - 2.6.2
+#          - 2.6.3
+#          - 2.7.1
+#          - 2.7.2
+#          - 2.8.1
+#          - master
```
Totktonada, line 166: Use `-DENABLE_DIST=ON` cmake option to install tarantoolctl into a destination directory.

test/integration/hotreload_test.lua
```diff
@@ -0,0 +1,98 @@
+local fio = require('fio')
```
Totktonada, line 1: What property of expirationd you verify here? I don't understand how all this code is relevant to expirationd.

.github/workflows/test_on_push.yaml
Totktonada: 'This is small' does not advocate mixing of irrelevant changes into one commit. It has no any sense except saving 10 minutes of `git reset HEAD^ && git add -p` for you. If you don't want to pay this time to present you code in a good shape, why should I pay my time to review it?

test/integration/hotreload_test.lua
Totktonada: There should be a well defined hypothesis about a class of problems we can catch this way (plus understanding why we can't do it in a more obvious and simple way) to actually add and support this testing code within the expirationd repository.

Totktonada: Please, take me right: I encourage using of smart testing techniques, where usual unit / functional testing are not powerful enough. Say, when a functional testing is able to catch particular set of highly specialized problems, but well implemented integration testing is able to catch almost any problem of given class. Or, say, the typical situation, when a functional testing unable to catch a real world problem, which only appears with data from a previous version of a module / an application.

I understand what cartridge/hotreload.lua doing (as well as moonlibs/module-reload). But I still don't catch the idea of this test. The questions that I ask for myself are:

* What is the class of problems it should catch?
* Whether it is beyond the class of problems that will be catched by the following code:

  ```lua

  <..create a space..>
  <..fill the space..>
  expirationd.start(<...>)
  <..check all tuples are expired..>

  expirationd.kill(<...>)
  <..fill the space..>
  <..check that tuples are not expiring..>

  package.loaded.expirationd = nil

  expirationd = require('expirationd')
  expirationd.start(<...>)
  <..check all tuples are expired..>
  ```
* Even more, whether it is beyoud problems we able to catch this way:

  ```lua
  require('expirationd')
  package.loaded.expirationd = nil
  require('expirationd')
  ```

.github/workflows/test_on_push.yaml
ArtDu: Just a little weird to add ci to testing without a single test, but okay: https://github.com/tarantool/expirationd/pull/55

.github/workflows/testing.yaml
ArtDu: My mistake is that I did not mark this PR as draft / WIP

.github/workflows/testing.yaml
ArtDu: Also we have to add LUA_PATH to ENV and pass TARANTOOL_DIR to tarantoolctl rocks install, but there will be problems like we did when installing cartridge(http), see https://github.com/tarantool/expirationd/blob/9edcbfd10f77680befdaf4b2a49b115702f2e7d9/.github/workflows/testing.yaml#L223-L224

test/integration/hotreload_test.lua
ArtDu: in this case, @no1seman asked to make a test, which will clearly show that when call require("cartridge.roles").reload(), correct processing of the expirationd module is possible

.github/workflows/test_on_push.yaml
ArtDu: moved it to another PR #55

.github/workflows/testing.yaml
ArtDu: moved it to another PR

.github/workflows/testing.yaml
ArtDu: moved it to another PR #55

vrogach2020:
- split test cases into unit tests and integration tests
- keep unit tests in this repo
- if integration tests will require cartridge dependency probably we will move them to other repo (checking bundle integrations)

ligurio:
> After discussion with @ligurio and @ArtDu there are some follow-ups:

So I'll close PR. Reopen it when changes will be ready to review.

Issue #50, opened 2021-04-15
RFC: Adding Iterations over Secondary Indexes 


ArtDu:
* **Issues**:  https://github.com/tarantool/expirationd/issues/48

## Problem
We want to use secondary indexes to iterate over the space, as is done in indexpiration, but also use all expirationd features such as callback and hotreload.

## Needs
* Make it available to specify the index for iteration
* It is necessary to preserve backward compatibility, so that if the index is not specified, then the iteration goes as it is now on the primary key

### Iterations over Secondary
Now there is a hardcode implementation of iterations by tree or hash index zero, i.e. by primary:
https://github.com/tarantool/expirationd/blob/29d1a25962bb5633d094bf7b016480690226da14/expirationd.lua#L91-L100

The logic in **indexepiration** is now this, some field is used to remove by time (only the time or time64 types are available):
If the time value in the field is greater than zero, then we walk along it in the index:
```lua
for _,t in expire_index:pairs({0},{iterator = box.index.GT}) do
```
```lua
	if opts.kind == 'time' or opts.kind == 'time64' then
		if not typeeq(expire_index.parts[1].type,'num') then
			error(("Can't use field %s as %s"):format(opts.field,opts.kind),2+depth)
		end
		if opts.kind == 'time' then
			self.check = function(t)
				return t[ expire_field_no ] - clock.realtime()
			end
		elseif opts.kind == 'time64' then
			self.check = function(t)
				return tonumber(
					ffi.cast('int64_t',t[ expire_field_no ])
					- ffi.cast('int64_t',clock.realtime64())
				)/1e9
			end
		end
	elseif _callable(opts.kind) then
		self.check = opts.kind
	else
		error(("Unsupported kind: %s"):format(opts.kind),2+depth)
	end

```

In **expirationd** we can use the specified index from opts and perhaps we need to specify from which element to start the iteration like:
```lua
-- as noticed @akudiyar we need direction and stop iteration function
iterator = 'GE'
if not ascending then
  iterator = 'LE'
end
local params = {iterator = iterator, limit = task.tuples_per_iteration}
local tuples = scan_space.index.expire_index:select({start_element}, params)
while #tuples > 0 do
        last_id = tuples[#tuples]
        for _, tuple in ipairs(tuples) do
            if stop_iteration(task) then break end
            expiration_process(task, tuple)
        end
        suspend(scan_space, task)
        local key = construct_key(scan_space.id, last_id)
        -- select all greater then last key
[...]

Issue #38, opened 2019-07-30
How to run expirationd periodically?


santhoshTpixler:
I want to run the expirationd periodically without manually triggering it. For example, I want to run it once every hour. Is it possible?

Totktonada:
It seems, it worth to clarify how expirationd works. The schema is the following:

1. Process min(`space_length`, `tuples_per_iteration`) tuples.
2. Sleep `tuples_per_iteration` × `full_scan_time` / `space_length` (but not beyond 1 second).
3. Repeat 1-2 until the whole space will be traversed.
4. Sleep 1 second.
5. Repeat 1-4.

Example 1:

```
For given values:

 | space_length = 10^6
 | tuples_per_iteration = 1024 (default)
 | full_scan_time = 3600 (default, 1 hour)

We can calculate delay = min(1024 * 3600 / 10^6, 1) = 1.

The processing will look so:

+-> process 1024 tuples
|   sleep 1 second
|   <..975 more such steps..>
|   process 576 remaining tuples
|   sleep 1
|   sleep 1
|   (978 seconds of delays in sum, ~16 minutes)
+-- start again
```

Example 2:

```
For given values:

 | space_length = 10^6
 | tuples_per_iteration = 1
 | full_scan_time = 3600 (default, 1 hour)

We can calculate delay = min(3600 / 10^6, 1) = 0.0036.

The processing will look so:

+-> process 1 tuple
|   sleep 0.0036 seconds
|   <..10^6-1 more such steps..>
|   sleep 1
|   (3601 seconds of delays in sum, ~1 hour)
+-- start again
```

I think everything will be less obscure if we'll add ability to control maximum sleep time after processing of `tuple_per_iteration` tuples (now the maximum is 1 second) and to control a sleep time after traversing the whole space (it is hardcoded to 1 second). I hope we'll do it in PR #19. When it will be done, I'll return here with some recipes.

akudiyar:
Need to add the comment https://github.com/tarantool/expirationd/issues/38#issuecomment-715667451 into README

Totktonada:
NB: Don't forget to add `full_scan_delay` into the processing schema and the examples above, when we'll adopt them for README.

PR #52, opened 2021-04-20
Features for more customized behavior

ArtDu:
* Iterations over Secondary Indexes - name of the index to iterate on
* Transaction support - ability to process tuples from each batch in a single transaction
* Start key - start iterating from the tuple with this index value
* Iterator type - type of the iterator to use, as string or box.index constant
* Process while function - function to call before checking each tuple, if it returns false, the task will stop until next full scan
* And the ability to write custom iterator behavior for the user - `iterate_with` function which returns an iterator object which provides tuples to check

Additional changes:
Rewritten space iteration for tree index from select to pairs since the iterator is now stable
Added some comments and removed duplicate code(expirationd_kill_task)
Described new parameters in the code

Test coverage added. 

RFC: #50

### DIFF
--- a/.github/workflows/fast_testing.yaml
+++ b/.github/workflows/fast_testing.yaml
@@ -35,4 +35,16 @@ jobs:
       - name: Clone the module
         uses: actions/checkout@v2
 
+      - name: Cache rocks
+        uses: actions/cache@v2
+        id: cache-rocks
+        with:
+          path: .rocks/
+          key: cache-rocks-${{ matrix.tarantool }}-01
+
+      - name: Install requirements
+        run: |
+          tarantoolctl rocks install luatest 0.5.0
+        if: steps.cache-rocks.outputs.cache-hit != 'true'
+
       - run: make test
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
 all:
 	@echo "Only tests are available: make test"
 
+.PHONY: test
 test:
+	.rocks/bin/luatest -v
 	rm -rf *.xlog* *.snap 51{2,3,4,5,6,7}
 	INDEX_TYPE='TREE' SPACE_TYPE='vinyl' ./test.lua
 	rm -rf *.xlog* *.snap 51{2,3,4,5,6,7}
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ Run a scheduled task to check and process (expire) tuples in a given space.
      The index value may be a single value, if the index consists of one field, a tuple with the index key parts, or a function which returns such value.
      If omitted or nil, all tuples will be checked.
     * `tuples_per_iteration` - Number of tuples to check in one batch (iteration). Default is 1024.
+    * `atomic_iteration` - Boolean, false (default) to process each tuple as a single transaction.
+     True to process tuples from each batch in a single transaction.
     * `process_while` - Function to call before checking each tuple.
      If it returns false, the current tuple scan task finishes.
     * `iterate_with` - Function which returns an iterator object which provides tuples to check, considering the start_key, process_while and other options.
--- a/README.md
+++ b/README.md
@@ -50,6 +50,10 @@ Run a scheduled task to check and process (expire) tuples in a given space.
      The index value may be a single value, if the index consists of one field, a tuple with the index key parts, or a function which returns such value.
      If omitted or nil, all tuples will be checked.
     * `tuples_per_iteration` - Number of tuples to check in one batch (iteration). Default is 1024.
+    * `process_while` - Function to call before checking each tuple.
+     If it returns false, the current tuple scan task finishes.
+    * `iterate_with` - Function which returns an iterator object which provides tuples to check, considering the start_key, process_while and other options.
+     There's a default function which can be overriden with this parameter.
     * `on_full_scan_start` - Function to call before starting a tuple scan.
     * `on_full_scan_complete` - Function to call after completing a full scan.
     * `on_full_scan_success` - Function to call after successfully completing a full scan.
--- a/README.md
+++ b/README.md
@@ -13,8 +13,8 @@ being deleted into some other space.
 ``` lua
 box.cfg{}
 space = box.space.old
-job_name = 'clean_all'
-expirationd = require('expirationd')
+job_name = "clean_all"
+expirationd = require("expirationd")
 function is_expired(args, tuple)
   return true
 end
@@ -31,35 +31,30 @@ expirationd.start(job_name, space.id, is_expired, {
 
 ### `expirationd.start (name, space_id, is_tuple_expired, options)`
 
-Run a named task
+Run a scheduled task to check and process (expire) tuples in a given space.
 
 * `name` - task name
 * `space_id` - space to look in for expired tuples
 * `is_tuple_expired` - a function, must accept tuple and return true/false
   (is tuple expired or not), receives `(args, tuple)` as arguments
-opt
 * `options` -- (table with named options, may be nil)
-  * `process_expired_tuple` - applied to expired tuples, receives `(space_id, args, tuple)`
-    as arguments. Can be nil: by default tuples are removed
-  * `args` - passed to `is_tuple_expired()` and `process_expired_tuple()` as additional context
-  * `tuples_per_iteration` - number of tuples will be checked by one iteration
-  * `full_scan_time` - time required for full index scan (in seconds)
-  * `iteration_delay` - max sleep time between iterations (in seconds)
-  * `full_scan_delay` - sleep time between full scans (in seconds)
-  * `on_full_scan_start` - call function on starting full scan iteration
-    Receives `(task)` as arguments.
-  * `on_full_scan_complete` - call function on complete full scan iteration.
-    Called after `on_full_scan_success` or `on_full_scan_error`.
-    Receives `(task)` as arguments.
-  * `on_full_scan_success` - call function on success full scan iteration
-    Receives `(task)` as arguments.
-  * `on_full_scan_error` - call function on error full scan iteration
-    Receives `(task, error)` as arguments.
-  * `force` - run, even on replica
+    * `process_expired_tuple` - Applied to expired tuples, receives (space_id, args, tuple) as arguments.
+     Can be nil: by default, tuples are removed.
+    * `tuples_per_iteration` - Number of tuples to check in one batch (iteration). Default is 1024.
+    * `on_full_scan_start` - Function to call before starting a tuple scan.
+    * `on_full_scan_complete` - Function to call after completing a full scan.
+    * `on_full_scan_success` - Function to call after successfully completing a full scan.
+    * `on_full_scan_error` - Function to call after terminating a full scan due to an error.
+    * `args` - Passed to is_tuple_expired and process_expired_tuple() as an additional context.
+    * `full_scan_time` - Time required for a full index scan (in seconds).
+    * `iteration_delay` - Max sleep time between batches (in seconds).
+    * `full_scan_delay` - Sleep time between full scans (in seconds).
+    * `force` - Run task even on replica.
+
 
 ### `expirationd.kill (name)`
 
-Kill an existing task with name 'name'
+Kill an existing task with name "name"
 
 * `name` - task's name
 
--- a/README.md
+++ b/README.md
@@ -43,6 +43,12 @@ Run a scheduled task to check and process (expire) tuples in a given space.
     * `index` - Name or id of the index to iterate on. If omitted, will use the primary index.
      If there's no index with this name, will throw an error.
      Supported index types are TREE and HASH, using other types will result in an error.
+    * `iterator_type` - Type of the iterator to use, as string or box.index constant, for example, "EQ" or box.index.EQ.
+     Default is box.index.ALL.
+     See https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_index/pairs/.
+    * `start_key` - Start iterating from the tuple with this index value. If the iterator is "EQ", iterate over tuples with this index value.
+     The index value may be a single value, if the index consists of one field, a tuple with the index key parts, or a function which returns such value.
+     If omitted or nil, all tuples will be checked.
     * `tuples_per_iteration` - Number of tuples to check in one batch (iteration). Default is 1024.
     * `on_full_scan_start` - Function to call before starting a tuple scan.
     * `on_full_scan_complete` - Function to call after completing a full scan.
--- a/README.md
+++ b/README.md
@@ -9,8 +9,10 @@ tuple and performs the expiry itself: either deletes it (memcache), or
 does something smarter, like put a smaller representation of the data
 being deleted into some other space.
 
-### Example
-``` lua
+### Examples
+
+Simple version:
+```lua
 box.cfg{}
 space = box.space.old
 job_name = "clean_all"
@@ -24,6 +26,44 @@ end
 expirationd.start(job_name, space.id, is_expired, {
     process_expired_tuple = delete_tuple, args = nil,
     tuples_per_iteration = 50, full_scan_time = 3600
+})
+```
+
+Сustomized version:
+```lua
+expirationd.start(job_name, space.id, is_expired, {
+    -- name or id of the index in the specified space to iterate over
+    index = "exp",
+    -- one transaction per batch
+    -- default is false
+    atomic_iteration = true,
+    -- delete data that was added a year ago
+    -- default is nil
+    start_key = function( task )
+        return clock.time() - (365*24*60*60)
+    end,
+    -- delete it from the oldest to the newest
+    -- default is ALL
+    iterator_type = "GE",
+    -- stop full_scan if delete a lot
+    -- returns true by default
+    process_while = function( task )
+        if task.args.max_expired_tuples >= task.expired_tuples_count then
+            task.expired_tuples_count = 0
+            return false
+        end
+        return true
+    end,
+    -- this function must return an iterator over the tuples
+    iterate_with = function( task )
+        return task.expire_index:pairs({ task.start_key() }, { iterator = task.iterator })
+            :take_while( function( tuple )
+                return task:process_while()
+            end )
+    end,
+    args = {
+        max_expired_tuples = 1000
+    }
 })
 ```
 
@@ -112,4 +152,4 @@ Get statistics of task
 
 ## Testing
 
-Simply start `tarantool test.lua`
+Simply start `make test`
--- a/README.md
+++ b/README.md
@@ -40,6 +40,9 @@ Run a scheduled task to check and process (expire) tuples in a given space.
 * `options` -- (table with named options, may be nil)
     * `process_expired_tuple` - Applied to expired tuples, receives (space_id, args, tuple) as arguments.
      Can be nil: by default, tuples are removed.
+    * `index` - Name or id of the index to iterate on. If omitted, will use the primary index.
+     If there's no index with this name, will throw an error.
+     Supported index types are TREE and HASH, using other types will result in an error.
     * `tuples_per_iteration` - Number of tuples to check in one batch (iteration). Default is 1024.
     * `on_full_scan_start` - Function to call before starting a tuple scan.
     * `on_full_scan_complete` - Function to call after completing a full scan.
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -14,9 +14,9 @@
 -- local support functions
 -- ========================================================================= --
 
-local fun = require('fun')
-local log = require('log')
-local fiber = require('fiber')
+local fun = require("fun")
+local log = require("log")
+local fiber = require("fiber")
 
 -- get fiber id function
 local function get_fiber_id(fiber)
@@ -45,7 +45,7 @@ local constants = {
     -- factor for recalculation of vinyl space size
     default_vinyl_assumed_space_len_factor = 2,
     -- default function on full scan
-    default_on_full_scan = function(...) end
+    default_on_full_scan = function() end,
 }
 
 -- ========================================================================= --
@@ -56,6 +56,7 @@ local constants = {
 -- Task fibers
 -- ------------------------------------------------------------------------- --
 
+-- get all fields in primary key(composite possible) from tuple
 local function construct_key(space_id, tuple)
     return fun.map(
         function(x) return tuple[x.fieldno] end,
@@ -63,6 +64,7 @@ local function construct_key(space_id, tuple)
     ):totable()
 end
 
+-- do expiration process on tuple
 local function expiration_process(task, tuple)
     task.checked_tuples_count = task.checked_tuples_count + 1
     if task.is_tuple_expired(task.args, tuple) then
@@ -71,6 +73,7 @@ local function expiration_process(task, tuple)
     end
 end
 
+-- yield for some time
 local function suspend_basic(scan_space, task, len)
     local delay = (task.tuples_per_iteration * task.full_scan_time)
     delay = math.min(delay / len, task.iteration_delay)
@@ -78,6 +81,7 @@ local function suspend_basic(scan_space, task, len)
 end
 
 local function suspend(scan_space, task)
+    -- Return the number of tuples in the space
     local space_len = scan_space:len()
     if space_len > 0 then
         suspend_basic(scan_space, task, space_len)
@@ -86,7 +90,7 @@ end
 
 local function tree_index_iter(scan_space, task)
     -- iteration with GT iterator
-    local params = {iterator = 'GT', limit = task.tuples_per_iteration}
+    local params = {iterator = "GT", limit = task.tuples_per_iteration}
     local last_id
     local tuples = scan_space.index[0]:select({}, params)
     while #tuples > 0 do
@@ -120,7 +124,7 @@ local function default_do_worker_iteration(task)
     local index_type = scan_space.index[0].type
 
     -- full index scan loop
-    if index_type == 'HASH' then
+    if index_type == "HASH" then
         hash_index_iter(scan_space, task)
     else
         tree_index_iter(scan_space, task)
@@ -133,7 +137,7 @@ local function vinyl_do_worker_iteration(task)
     local checked_tuples_count = 0
     local space_len = task.vinyl_assumed_space_len
 
-    local params = {iterator = 'GT', limit = task.tuples_per_iteration}
+    local params = {iterator = "GT", limit = task.tuples_per_iteration}
     local tuples = scan_space.index[0]:select({}, params)
     while true do
         local tuple_cnt = #tuples
@@ -186,6 +190,7 @@ local function guardian_loop(task)
     fiber.name(string.format("guardian of %q", task.name), { truncate = true })
 
     while true do
+        -- if fiber doesn't exist
         if get_fiber_id(task.worker_fiber) == 0 then
             -- create worker fiber
             task.worker_fiber = fiber.create(worker_loop, task)
@@ -205,7 +210,7 @@ end
 -- * task:start()      -- start task
 -- * task:stop()       -- stop task
 -- * task:restart()    -- restart task
--- * task:kill()       -- delete task and restart
+-- * task:kill()       -- stop task and delete
 -- * task:statistics() -- return table with statistics
 local Task_methods = {
     start = function (self)
@@ -215,14 +220,14 @@ local Task_methods = {
     stop = function (self)
         if (get_fiber_id(self.guardian_fiber) ~= 0) then
             self.guardian_fiber:cancel()
-            while self.guardian_fiber:status() ~= 'dead' do
+            while self.guardian_fiber:status() ~= "dead" do
                 fiber.sleep(0.01)
             end
             self.guardian_fiber = nil
         end
         if (get_fiber_id(self.worker_fiber) ~= 0) then
             self.worker_fiber:cancel()
-            while self.worker_fiber:status() ~= 'dead' do
+            while self.worker_fiber:status() ~= "dead" do
                 fiber.sleep(0.01)
             end
             self.worker_fiber = nil
@@ -269,7 +274,7 @@ local function create_task(name)
         on_full_scan_error             = constants.default_on_full_scan,
         on_full_scan_success           = constants.default_on_full_scan,
         on_full_scan_start             = constants.default_on_full_scan,
-        on_full_scan_complete          = constants.default_on_full_scan
+        on_full_scan_complete          = constants.default_on_full_scan,
     }, { __index = Task_methods })
     return task
 end
@@ -298,7 +303,8 @@ end
 -- Expiration daemon management functions
 -- ========================================================================= --
 
--- Run a named task
+-- Run a scheduled task to check and process (expire) tuples in a given space.
+--
 -- params:
 --   name             -- task name
 --   space_id         -- space to look in for expired tuples
@@ -306,19 +312,19 @@ end
 --                       true/false (is tuple expired or not),
 --                       receives (args, tuple) as arguments
 --   options = {      -- (table with named options)
---     * process_expired_tuple -- applied to expired tuples, receives
---                                (space_id, args, tuple) as arguments
---     * on_full_scan_start    -- call function on starting full scan iteration
---     * on_full_scan_complete -- call function on complete full scan iteration
---     * on_full_scan_success  -- call function on success full scan iteration
---     * on_full_scan_error    -- call function on error full scan iteration
---     * args                  -- passed to is_tuple_expired and
---                                process_expired_tuple() as additional context
---     * tuples_per_iteration  -- number of tuples will be checked by one iteration
---     * full_scan_time        -- time required for full index scan (in seconds)
---     * iteration_delay       -- max sleep time between iterations (in seconds)
---     * full_scan_delay       -- sleep time between full scans (in seconds)
---     * force                 -- run task even on replica
+--     * process_expired_tuple -- Applied to expired tuples, receives (space_id, args, tuple) as arguments;
+--                                can be nil: by default, tuples are removed.
+--     * on_full_scan_start    -- Function to call before starting a full scan iteration.
+--     * on_full_scan_complete -- Function to call after completing a full scan iteration.
+--     * on_full_scan_success  -- Function to call after successfully completing a full scan iteration.
+--     * on_full_scan_error    -- Function to call after terminating a full scan due to an error.
+--     * args                  -- Passed to is_tuple_expired and
+--                                process_expired_tuple() as additional context.
+--     * tuples_per_iteration  -- Number of tuples to check in one batch (iteration); default is 1024.
+--     * full_scan_time        -- Time required for a full index scan (in seconds).
+--     * iteration_delay       -- Max sleep time between iterations (in seconds).
+--     * full_scan_delay       -- Sleep time between full scans (in seconds).
+--     * force                 -- Run task even on replica.
 --  }
 local function expirationd_run_task(name, space_id, is_tuple_expired, options)
     if name == nil then
@@ -347,7 +353,7 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
     task.is_tuple_expired = is_tuple_expired
 
     -- optional params
-    if options ~= nil and type(options) ~= 'table' then
+    if options ~= nil and type(options) ~= "table" then
         error("options must be table or not defined")
     end
     options = options or {}
@@ -365,7 +371,7 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
     -- check tuples per iteration (not required)
     if options.tuples_per_iteration ~= nil then
         if options.tuples_per_iteration <= 0 then
-            error("invalid tuples per iteration parameter")
+            error("Invalid tuples per iteration parameter")
         end
         task.tuples_per_iteration = options.tuples_per_iteration
     end
@@ -373,76 +379,76 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
     -- check full scan time
     if options.full_scan_time ~= nil then
         if options.full_scan_time <= 0 then
-            error("invalid full scan time")
+            error("Invalid full scan time")
         end
         task.full_scan_time = options.full_scan_time
     end
 
     if options.force ~= nil then
-        if type(options.force) ~= 'boolean' then
+        if type(options.force) ~= "boolean" then
             error("Invalid type of force value")
         end
         task.force = options.force
     end
 
     if options.vinyl_assumed_space_len ~= nil then
-        if type(options.vinyl_assumed_space_len) ~= 'number' then
+        if type(options.vinyl_assumed_space_len) ~= "number" then
             error("Invalid type of vinyl_assumed_space_len value")
         end
         task.vinyl_assumed_space_len = options.vinyl_assumed_space_len
     end
 
     if options.vinyl_assumed_space_len_factor ~= nil then
-        if type(options.vinyl_assumed_space_len_factor) ~= 'number' then
+        if type(options.vinyl_assumed_space_len_factor) ~= "number" then
             error("Invalid type of vinyl_assumed_space_len_factor value")
         end
         task.vinyl_assumed_space_len_factor = options.vinyl_assumed_space_len_factor
     end
 
-    if box.space[task.space_id].engine == 'vinyl' then
+    if box.space[task.space_id].engine == "vinyl" then
         task.do_worker_iteration = vinyl_do_worker_iteration
     else
         task.do_worker_iteration = default_do_worker_iteration
     end
 
     if options.iteration_delay ~= nil then
-        if type(options.iteration_delay) ~= 'number' then
-            error("invalid type of iteration_delay value")
+        if type(options.iteration_delay) ~= "number" then
+            error("Invalid type of iteration_delay value")
         end
         task.iteration_delay = options.iteration_delay
     end
 
     if options.full_scan_delay ~= nil then
-        if type(options.full_scan_delay) ~= 'number' then
-            error("invalid type of full_scan_delay value")
+        if type(options.full_scan_delay) ~= "number" then
+            error("Invalid type of full_scan_delay value")
         end
         task.full_scan_delay = options.full_scan_delay
     end
 
     if options.on_full_scan_start ~= nil then
-        if type(options.on_full_scan_start) ~= 'function' then
-            error("invalid type of on_full_scan_start is not function")
+        if type(options.on_full_scan_start) ~= "function" then
+            error("Invalid type of on_full_scan_start, expected function")
         end
         task.on_full_scan_start = options.on_full_scan_start
     end
 
     if options.on_full_scan_success ~= nil then
-        if type(options.on_full_scan_success) ~= 'function' then
-            error("invalid type of on_full_scan_success is not function")
+        if type(options.on_full_scan_success) ~= "function" then
+            error("Invalid type of on_full_scan_success, expected function")
         end
         task.on_full_scan_success = options.on_full_scan_success
     end
 
     if options.on_full_scan_complete ~= nil then
-        if type(options.on_full_scan_complete) ~= 'function' then
-            error("invalid type of on_full_scan_complete is not function")
+        if type(options.on_full_scan_complete) ~= "function" then
+            error("Invalid type of on_full_scan_complete, expected function")
         end
         task.on_full_scan_complete = options.on_full_scan_complete
     end
 
     if options.on_full_scan_error ~= nil then
-        if type(options.on_full_scan_error) ~= 'function' then
-            error("invalid type of on_full_scan_error is not function")
+        if type(options.on_full_scan_error) ~= "function" then
+            error("Invalid type of on_full_scan_error, expected function")
         end
         task.on_full_scan_error = options.on_full_scan_error
     end
@@ -502,10 +508,7 @@ local function expirationd_task_stats(name)
     return retval
 end
 
--- kill task
-local function expirationd_kill_task(name)
-    return get_task(name):kill()
-end
+
 
 -- get task by name
 local function expirationd_get_task(name)
@@ -517,15 +520,15 @@ end
 -- * require new expirationd
 -- * restart all tasks
 local function expirationd_update()
-    local expd_prev = require('expirationd')
+    local expd_prev = require("expirationd")
     table.clear(expd_prev)
     setmetatable(expd_prev, {
         __index = function(name)
             error("Wait until update is done before using expirationd", 2)
         end
     })
-    package.loaded['expirationd'] = nil
-    local expd_new  = require('expirationd')
+    package.loaded["expirationd"] = nil
+    local expd_new  = require("expirationd")
     local tmp_task_list = task_list; task_list = {}
     for name, task in pairs(tmp_task_list) do
         task:kill()
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -46,6 +46,10 @@ local constants = {
     default_vinyl_assumed_space_len_factor = 2,
     -- default function on full scan
     default_on_full_scan = function() end,
+    -- default function for start_key
+    start_key = function() return nil end,
+    -- default iterating over the loop will go in ascending index
+    iterator_type = "ALL",
 }
 
 -- ========================================================================= --
@@ -93,7 +97,7 @@ local function default_do_worker_iteration(task)
     local space_len = task.vinyl_assumed_space_len
     local checked_tuples_count = 0
     local vinyl_checked_tuples_count = 0
-    for _, tuple in task.index:pairs(nil, {iterator = box.index.ALL}) do
+    for _, tuple in task.index:pairs(task.start_key(), {iterator = task.iterator_type}) do
         checked_tuples_count = checked_tuples_count + 1
         vinyl_checked_tuples_count = vinyl_checked_tuples_count + 1
         expiration_process(task, tuple)
@@ -232,6 +236,8 @@ local function create_task(name)
         on_full_scan_success           = constants.default_on_full_scan,
         on_full_scan_start             = constants.default_on_full_scan,
         on_full_scan_complete          = constants.default_on_full_scan,
+        start_key                      = constants.start_key,
+        iterator_type                  = constants.iterator_type,
     }, { __index = Task_methods })
     return task
 end
@@ -274,13 +280,21 @@ end
 --     * index                 -- Name or id of the index to iterate on; if omitted, will use the primary index;
 --                                if there's no index with this name, will throw an error.
 --                                supported index types are TREE and HASH, using other types will result in an error.
+--     * iterator_type         -- Type of the iterator to use, as string or box.index constant,
+--                                for example, 'EQ' or box.index.EQ; default is box.index.ALL;
+--                                see https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_index/pairs/.
+--     * start_key             -- Start iterating from the tuple with this index value;
+--                                or when iterator is 'EQ', iterate over tuples with this index value;
+--                                must be a value of the same data type as the index field or fields,
+--                                or a function which returns such value;
+--                                if omitted or nil, all tuples will be checked.
+--     * tuples_per_iteration  -- Number of tuples to check in one batch (iteration); default is 1024.
 --     * on_full_scan_start    -- Function to call before starting a full scan iteration.
 --     * on_full_scan_complete -- Function to call after completing a full scan iteration.
 --     * on_full_scan_success  -- Function to call after successfully completing a full scan iteration.
 --     * on_full_scan_error    -- Function to call after terminating a full scan due to an error.
 --     * args                  -- Passed to is_tuple_expired and
 --                                process_expired_tuple() as additional context.
---     * tuples_per_iteration  -- Number of tuples to check in one batch (iteration); default is 1024.
 --     * full_scan_time        -- Time required for a full index scan (in seconds).
 --     * iteration_delay       -- Max sleep time between iterations (in seconds).
 --     * full_scan_delay       -- Sleep time between full scans (in seconds).
@@ -343,6 +357,23 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
         end
     end
     task.index = expire_index
+
+    -- check iterator_type
+    if options.iterator_type ~= nil then
+        task.iterator_type = options.iterator_type
+    end
+
+    -- check start_key
+    if options.start_key ~= nil or options.start_key == box.NULL then
+        if type(options.start_key) == "function" then
+            task.start_key = function() return options.start_key() end
+        else
+            task.start_key = function() return options.start_key end
+        end
+    end
+
+    -- check valid of iterator_type and start key
+    task.index:pairs( task.start_key(), { iterator = task.iterator_type })
 
     -- check expire and process after expiration handler's arguments
     task.args = options.args
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -52,6 +52,8 @@ local constants = {
     process_while = function() return true end,
     -- default iterating over the loop will go in ascending index
     iterator_type = "ALL",
+    -- default atomic_iteration is false, batch of items doesn't include in one transaction
+    atomic_iteration = false,
 }
 
 -- ========================================================================= --
@@ -99,6 +101,9 @@ local function default_do_worker_iteration(task)
     local space_len = task.vinyl_assumed_space_len
     local checked_tuples_count = 0
     local vinyl_checked_tuples_count = 0
+    if task.atomic_iteration then
+        box.begin()
+    end
     for _, tuple in task:iterate_with() do
         checked_tuples_count = checked_tuples_count + 1
         vinyl_checked_tuples_count = vinyl_checked_tuples_count + 1
@@ -106,6 +111,12 @@ local function default_do_worker_iteration(task)
         -- find out if the worker can go to sleep
         -- if the batch is full
         if checked_tuples_count >= task.tuples_per_iteration then
+            if task.atomic_iteration then
+                box.commit()
+                if task.worker_canceled then
+                    return true
+                end
+            end
             checked_tuples_count = 0
             if box.space[task.space_id].engine == "vinyl" then
                 if vinyl_checked_tuples_count > space_len then
@@ -115,7 +126,13 @@ local function default_do_worker_iteration(task)
             else
                 suspend(task)
             end
+            if task.atomic_iteration then
+                box.begin()
+            end
         end
+    end
+    if task.atomic_iteration then
+        box.commit()
     end
     if box.space[task.space_id].engine == "vinyl" then
         task.vinyl_assumed_space_len = vinyl_checked_tuples_count
@@ -127,9 +144,15 @@ local function worker_loop(task)
     fiber.name(string.format("worker of %q", task.name), { truncate = true })
 
     while true do
+        if task.worker_canceled then
+            fiber.self():cancel()
+        end
         if (box.cfg.replication_source == nil and box.cfg.replication == nil) or task.force then
             task.on_full_scan_start(task)
             local state, err = pcall(task.do_worker_iteration, task)
+            if task.worker_canceled then
+                fiber.self():cancel()
+            end
             if state then
                 task.on_full_scan_success(task)
             else
@@ -138,6 +161,7 @@ local function worker_loop(task)
 
             task.on_full_scan_complete(task)
             if not state then
+                box.rollback()
                 error(err)
             end
         end
@@ -188,7 +212,11 @@ local Task_methods = {
             self.guardian_fiber = nil
         end
         if (get_fiber_id(self.worker_fiber) ~= 0) then
-            self.worker_fiber:cancel()
+            if self.atomic_iteration then
+                self.worker_canceled = true
+            else
+                self.worker_fiber:cancel()
+            end
             while self.worker_fiber:status() ~= "dead" do
                 fiber.sleep(0.01)
             end
@@ -229,6 +257,7 @@ local function create_task(name)
         args                  = nil,
         index                 = nil,
         iterate_with          = nil,
+        worker_canceled       = false,
         iteration_delay                = constants.max_delay,
         full_scan_delay                = constants.max_delay,
         tuples_per_iteration           = constants.default_tuples_per_iteration,
@@ -242,6 +271,7 @@ local function create_task(name)
         start_key                      = constants.start_key,
         process_while                  = constants.process_while,
         iterator_type                  = constants.iterator_type,
+        atomic_iteration               = constants.atomic_iteration,
     }, { __index = Task_methods })
     return task
 end
@@ -303,6 +333,8 @@ end
 --                                or a function which returns such value;
 --                                if omitted or nil, all tuples will be checked.
 --     * tuples_per_iteration  -- Number of tuples to check in one batch (iteration); default is 1024.
+--     * atomic_iteration      -- Boolean, false (default) to process each tuple as a single transaction;
+--                                true to process tuples from each batch in a single transaction.
 --     * process_while         -- Function to call before checking each tuple;
 --                                if it returns false, the task will stop until next full scan.
 --     * iterate_with          -- Function which returns an iterator object which provides tuples to check,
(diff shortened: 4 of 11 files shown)


expirationd.lua
```diff
@@ -359,21 +384,131 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
...
+                    error("Wrong type in " .. i ..
+                             " element of start_element table, expected: " .. expire_index.parts[i].type ..
+                             ", got: " .. type(options.start_element[i]))
+                end
+            end
+        else
+            if #expire_index.parts ~= 1 then
+                error("Wrong number of elements in start_element table, expected: " .. #expire_index.parts)
+            end
+            if expire_index.parts[1].type ~= type(options.start_element) then
```
ArtDu, line 482: There will be an error if the index type is unsigned, and in options we accept number e.g. 123

test/unit/new_features_test.lua
```diff
@@ -0,0 +1,304 @@
...
+            table.insert(iteration_result, tuple)
+            return true
+        end
+        space:insert({1, "3"})
+        space:insert({2, "2"})
+        space:insert({3, "1"})
+
+        local task = expirationd.start("clean_all", space.id, is_expired)
+        -- wait for tuples expired
+        fiber.sleep(1)
```
ArtDu, line 175: maybe better use `helpers.retrying({}, function()`

expirationd.lua
Totktonada: As well for 'scalar' index part type, 'decimal'.

expirationd.lua
```diff
@@ -56,85 +58,90 @@ local constants = {
...
+            if task:stop_iteration() then goto done end
             expiration_process(task, tuple)
         end
-        suspend(scan_space, task)
-        local key = construct_key(scan_space.id, last_id)
-        tuples = scan_space.index[0]:select(key, params)
+        suspend(task)
+        local key = construct_key(task.expire_index, last_id)
+        -- select all greater then last key
+        tuples = task.expire_index:select(key, params)
```
olegrok, line 108: Approach with select won't work for non-unique secondary indexes.
See https://github.com/tarantool/tarantool/issues/3898 for details.

Single allowed you option is to use pairs here - https://github.com/tarantool/expirationd/pull/40

Test case is quite simple, real and painful. Insert `tuples_per_iteration + 1` with the same secondary index value. Then if you use 'GT' iterator you skip such records if you use 'GE' iterator it could be an infinite loop.

expirationd.lua
```diff
@@ -56,106 +62,80 @@ local constants = {
...
     ):totable()
 end
 
+-- do expiration process on tuple
 local function expiration_process(task, tuple)
     task.checked_tuples_count = task.checked_tuples_count + 1
     if task.is_tuple_expired(task.args, tuple) then
         task.expired_tuples_count = task.expired_tuples_count + 1
-        task.process_expired_tuple(task.space_id, task.args, tuple)
+        task.process_expired_tuple(task.space_id, task.args, tuple, task)
```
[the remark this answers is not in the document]
ArtDu: This will break backward compatibility. in previous versions of expirationd into a function that is not set by the user expiration_process call user defined function like `task.process_expired_tuple (task.space_id, task.args, tuple)`. It follows that the user-defined function used this data for the input parameters.

expirationd.lua
```diff
@@ -289,11 +276,22 @@ local function get_task(name)
 end
 
 -- default process_expired_tuple function
-local function default_tuple_drop(space_id, args, tuple)
-    box.space[space_id]:delete(construct_key(space_id, tuple))
+-- task as parameter was added to preserve backward compatibility
```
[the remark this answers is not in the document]
ArtDu: From this https://github.com/tarantool/expirationd/pull/52#discussion_r631687664 it follows that we should leave the first two input parameters, although they are not used, since expiration_process passes them to this function.

expirationd.lua
```diff
@@ -260,6 +241,9 @@ local function create_task(name)
         is_tuple_expired      = nil,
         process_expired_tuple = nil,
         args                  = nil,
+        expire_index          = nil,
```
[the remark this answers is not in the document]
ArtDu: This is a field that is not passed through the public api. It is only used in internal functions. Why do we need to change option descriptions?

expirationd.lua
```diff
@@ -359,21 +363,119 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
...
+        elseif expire_index.type == 'HASH' then
+            if  options.iterator ~= 'ALL' and
+                    options.iterator ~= 'EQ'  and
+                    options.iterator ~= 'GT'  then
+                error("Invalid iterator for HASH index, expected {ALL, EQ, GT}")
+            end
+        end
+        task.iterator = options.iterator
+    else
+        if expire_index.type == 'HASH' then
```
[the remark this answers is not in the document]
ArtDu: There is a check for valid iterators for the hash index, since, according to the documentation, we can use three iterators ALL, EQ, GT

expirationd.lua
```diff
@@ -359,21 +363,119 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
...
+            task.iterator = "ALL"
+        end
+    end
+
+    -- check start_key
+    if options.start_key then
+        if type(options.start_key) == "table" then
+            if #expire_index.parts ~= #options.start_key then
```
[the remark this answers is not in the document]
ArtDu: if you're talking about what if the index, for example, consists of two fields and we can request data like this `select ({key1, key2})` and so `select ({key1})`? Maybe remove the type checking altogether and leave it to the user's conscience?

expirationd.lua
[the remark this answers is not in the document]
olegrok: You can validate iterator just call `index:pairs(key, iterator)`. Of course your comparison is incorrect since it's possible to use numeric value from `box.index.` e.g. `box.index.EQ`.

Why `index:pairs(key, iterator)` will work. Because this command only creates iterator but doesn't touch space anyhow. I hope you remember (or could read in Alexey's email about iterator stability) that iterators are lazy.
And if you can create iterator then it's correct.

expirationd.lua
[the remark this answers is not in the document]
akudiyar: We'd better have consistent naming everywhere.

expirationd.lua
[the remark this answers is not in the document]
akudiyar: I think we can look for the type checking code in other modules like `tarantool/crud` and borrow it from there. The problem that I pointed out is that the actual key length is less or equal to the index key length and may be equal to 0.

expirationd.lua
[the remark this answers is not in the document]
akudiyar: Still don't get why we need to keep the old parameter list in the private function.

test/unit/new_features_test.lua
```diff
@@ -0,0 +1,432 @@
...
+    t.assert_equals(task.iterator, "ALL")
+    task:kill()
+
+    task = expirationd.start("clean_all", g.hash.id, is_expired_true, {iterator = "ALL"})
+    t.assert_equals(task.iterator, "ALL")
+    task:kill()
+
+    task = expirationd.start("clean_all", g.hash.id, is_expired_true, {iterator = "EQ"})
+    t.assert_equals(task.iterator, "EQ")
+    task:kill()
```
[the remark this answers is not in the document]
akudiyar: Probably, we need to file a ticket to the doc

expirationd.lua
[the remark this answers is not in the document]
akudiyar: Question is outdated after verbal discussion

expirationd.lua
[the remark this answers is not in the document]
ArtDu: Made the check by calling index: pairs (key, iterator), as suggested by @olegrok

expirationd.lua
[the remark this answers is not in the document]
ArtDu: Made the check by calling index: pairs (key, iterator), as suggested by @olegrok

expirationd.lua
ArtDu: Made the check by calling `index:pairs (key, iterator)`, as suggested by @olegrok

expirationd.lua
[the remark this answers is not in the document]
ArtDu: suggest a name) expire_index name was stolen from indexpiration

expirationd.lua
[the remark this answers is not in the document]
Totktonada: NB: There is `box.internal.check_iterator_type(opts, key_is_nil)`. Example: https://github.com/tarantool/graphql.0/pull/260/files#diff-29d7b710a75ba3836cc3266be5b3ba6db6e20757047f54414b5c91e7ccb2e607R29-R37

expirationd.lua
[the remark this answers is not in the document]
ArtDu: renamed to index, where possible, now there is an index, which is the name of the index, which is set by the user and the internal index, which implies the physical index of tarantool

rpm/tarantool-expirationd.spec
```diff
@@ -29,6 +31,10 @@ make test
 %install
 install -d %{buildroot}%{_datarootdir}/tarantool/
 install -m 0644 expirationd.lua %{buildroot}%{_datarootdir}/tarantool/
+curl -L https://tarantool.io/avEfjSi/release/1.10/installer.sh | bash
+yum install -y tarantool
+tarantoolctl rocks install luatest
+CMAKE_DUMMY_WEBUI=YES tarantoolctl rocks install cartridge
```
Totktonada, line 37: I doubt that integration tests should be run as part of package building process.

We can either skip them or setup a CI flow that will install the module from the package and run full testing suite.

rpm/tarantool-expirationd.spec
```diff
@@ -29,6 +31,10 @@ make test
 %install
 install -d %{buildroot}%{_datarootdir}/tarantool/
 install -m 0644 expirationd.lua %{buildroot}%{_datarootdir}/tarantool/
+curl -L https://tarantool.io/avEfjSi/release/1.10/installer.sh | bash
+yum install -y tarantool
```
Totktonada, line 35: 1. Use `rpm/prebuild.sh` file (and `debian/prebuild.sh` file) to prepare build environment. packpack calls it before rpmbuild / debuild. Using of a network within a package build script is generally discouraged.
2. You don't need to install tarantool explicitly, it is in the build dependencies.

rpm/tarantool-expirationd.spec
ArtDu: new tests will run in gh ci

rpm/tarantool-expirationd.spec
ArtDu: new tests will run in gh ci

README.md
```diff
@@ -27,35 +29,74 @@ expirationd.start(job_name, space.id, is_expired, {
...
+    * `start_key` - Start iterating from the tuple with this index value. Or when iterator is 'EQ', iterate over tuples with this index value. Must be a value of the same data type as the index field or fields, or a function which returns such value. If omitted or nil, all tuples will be checked.
+    * `tuples_per_iteration` - Number of tuples to check in one batch (iteration). Default is 1024.
+    * `atomic_iteration` - Boolean, false (default) to process each tuple as a single transaction. True to process tuples from each batch in a single transaction.
+    * `process_while` - Function to call before checking each tuple. If it returns false, the task will stop until next full scan.
+    * `iterate_with` - Function which returns an iterator object which provides tuples to check, considering the start_key, process_while and other options. There's a default function which can be overriden with this parameter.
+    * `on_full_scan_start` - Function to call before starting a full scan iteration.
```
akudiyar, line 90: before starting a tuple scan. 

I'd rather avoid mentioning "full scan" here and further, since it is not necessary full scan any more.

README.md
```diff
@@ -27,35 +29,74 @@ expirationd.start(job_name, space.id, is_expired, {
...
+    * `on_full_scan_complete` - Function to call after completing a full scan iteration.
+    * `on_full_scan_success` - Function to call after successfully completing a full scan iteration.
+    * `on_full_scan_error` - Function to call after terminating a full scan due to an error.
+    * `args` - Passed to is_tuple_expired and process_expired_tuple() as additional context.
```
akudiyar, line 94: as an additional context

README.md
```diff
@@ -27,35 +29,74 @@ expirationd.start(job_name, space.id, is_expired, {
...
+    * `full_scan_time` - Time required for a full index scan (in seconds).
```
akudiyar, line 95: This option is unclear -- does it allow to stop scanning after the specified period of time?

README.md
```diff
@@ -27,35 +29,74 @@ expirationd.start(job_name, space.id, is_expired, {
...
+    * `iteration_delay` - Max sleep time between iterations (in seconds).
```
akudiyar, line 96: Let's amend "iteration" to "batch" here and above.

README.md
```diff
@@ -27,35 +29,74 @@ expirationd.start(job_name, space.id, is_expired, {
...
 ### `expirationd.start (name, space_id, is_tuple_expired, options)`
 
-Run a named task
+Run a scheduled task to check and process (expire) tuples in a given space.
 
 * `name` - task name
 * `space_id` - space to look in for expired tuples
 * `is_tuple_expired` - a function, must accept tuple and return true/false
   (is tuple expired or not), receives `(args, tuple)` as arguments
 opt
```
akudiyar, line 80: A strange leftover?

README.md
```diff
@@ -27,35 +29,74 @@ expirationd.start(job_name, space.id, is_expired, {
...
+Сustomized version:
+```lua
+expirationd.start(job_name, space.id, is_expired, {
+    -- default is primary key
+    index = 'exp',
+    -- one transaction per batch
+    -- default is false
+    atomic_iteration = true,
+    -- delete data that was added a year ago
+    -- default return nil
```
akudiyar, line 41: default is nil

README.md
```diff
@@ -27,35 +29,74 @@ expirationd.start(job_name, space.id, is_expired, {
...
+    start_key = function( task )
+        return clock.time() - (365*24*60*60)
+    end,
+    -- delete it from the oldest to the newest
+    -- default is ALL
+    iterator_type = 'GE',
+    -- stop full_scan if delete a lot
+    -- default return true
```
akudiyar, line 49: returns true by default

README.md
```diff
@@ -27,35 +29,74 @@ expirationd.start(job_name, space.id, is_expired, {
...
+    process_while = function( task )
+        if task.args.max_expired_tuples >= task.expired_tuples_count then
+            task.expired_tuples_count = 0
+            return false
+        end
+        return true
+    end,
+    -- this function is default if no other is specified
```
akudiyar, line 57: this function must return an iterator over the tuples

Let's avoid saying that the listed function is "default" for not confusing users if the actual code will change in the future.

README.md
```diff
@@ -27,35 +29,74 @@ expirationd.start(job_name, space.id, is_expired, {
 })
 ```
 
+Сustomized version:
+```lua
+expirationd.start(job_name, space.id, is_expired, {
+    -- default is primary key
```
akudiyar, line 35: name of an index in the specified space to iterate over

expirationd.lua
```diff
@@ -359,21 +378,82 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
...
+        end
+    end
+
+    -- check valid of iterator_type and start key
+    task.index:pairs( task.start_key(), { iterator = task.iterator_type })
+
+    -- check process_while
+    if options.process_while ~= nil then
+        if type(options.process_while) ~= 'function' then
+            error("Invalid type of process_while is not function")
```
akudiyar, line 421: expected function

expirationd.lua
```diff
@@ -359,21 +378,82 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
...
+        if type(options.atomic_iteration) ~= 'boolean' then
+            error("Invalid type of atomic_iteration, expected boolean")
+        end
+        task.atomic_iteration = options.atomic_iteration
+    end
+
+    -- check iterate_with
+    if options.iterate_with ~= nil then
+        if type(options.iterate_with) ~= 'function' then
+            error("Invalid type of iterate_with is not function")
```
akudiyar, line 437: expected function

test/unit/atomic_iteration_test.lua
```diff
@@ -0,0 +1,110 @@
...
+
+function g.test_atomic_iteration()
+    helpers.iteration_result = {}
+
+    local transactions = {}
+    function f(iterator)
+        local transaction = {}
+        -- old / new_tuple is not passed for vinyl
+        for request_number, old_tuple, new_tuple, space_id in iterator() do
+            require'log'.info(old_tuple:totable())
```
akudiyar, line 33: Debug log?

akudiyar (changes requested):

README.md
ArtDu: I didn't quite understand what it meant. Write options instead opt?

README.md
ArtDu: There is a special ticket for this(https://github.com/tarantool/expirationd/issues/38). Perhaps a readme fix on this is needed by a separate PR. @Totktonada can you suggest a right way?

Totktonada: To be honest, I don't get Alexey's question.

.github/workflows/fast_testing.yaml
```diff
@@ -35,4 +35,23 @@ jobs:
...
+        if: steps.cache-rocks.outputs.cache-hit != 'true'
+
+      - name: Export path
+        run: echo $PWD/.rocks/bin >> $GITHUB_PATH
+
+      - name: Run luatest tests
+        run: luatest -v
+
+      - name: Run tap tests
+        run: make test
```
ligurio, line 57: For some reasons, you want to create new tests using `luatest` library. It's ok, but mixing different test libraries is not ok. Let's convert TAP tests to luatest's tests.

README.md
```diff
@@ -102,4 +153,5 @@ Get statistics of task
 
 ## Testing
 
-Simply start `tarantool test.lua`
+Tap tests `tarantool test.lua`  
+Luatests tests `luatest -v`
```
ligurio, line 157: We have a Makefile that does everything we need for testing, so it's enough to run command `make test` for testing expirationd.

test/unit/passing_optional_arguments_test.lua
```diff
@@ -0,0 +1,164 @@
...
+    t.assert_equals(task.iterator_type, "GT")
+    task:kill()
+
+    --errors
+    t.assert_error_msg_content_equals("Index 'primary' (HASH) of space 'hash' (memtx) does not support requested iterator type",
+            expirationd.start, "clean_all", g.hash.id, helpers.is_expired_true, {start_key = 1, iterator_type = 1})
+
+    t.assert_error_msg_content_equals("Index 'primary' (HASH) of space 'hash' (memtx) does not support requested iterator type",
+            expirationd.start, "clean_all", g.hash.id, helpers.is_expired_true, {start_key = 1, iterator_type = "GE"})
+
```
ligurio, line 99: I would split this huge testcase for number smaller testcases, one iterator type per testcase. When testcases are small it is much simpler to manage them.

ligurio (changes requested):
Artem, thanks for the patchset!

See my comments inline and I have general comments regarding the whole patchset.

We need to split patchset for smaller parts, especially patch "Features for more customized behavior", it mixes many different changes and it's become difficult to review them. I expect one commit per atomic change with clear reasons why this change is required. Please don't get me wrong, I don't want to slow down your work, I just want to follow best practices in software development.

I hope you already familiar with our Developer guidelines, if not you can always find recommendtions how to write a commit message in [1] and Lua style guide in [2].

1. https://www.tarantool.io/en/doc/latest/dev_guide/developer_guidelines/
2. https://www.tarantool.io/en/doc/latest/dev_guide/lua_style_guide/

expirationd.lua
```diff
@@ -252,8 +253,9 @@ local function get_task(name)
 end
 
 -- default process_expired_tuple function
-local function default_tuple_drop(space_id, args, tuple)
-    box.space[space_id]:delete(construct_key(space_id, tuple))
+-- task as parameter was added to preserve backward compatibility
+local function default_tuple_drop(space_id, args, tuple, task)
+    task.index:delete(construct_key(task.index, tuple))
```
olegrok, line 283: How will it work for non-unique indexes?

olegrok: Why isn't this change (in scope of this commit) covered with any tests?

expirationd.lua
```diff
@@ -322,6 +327,26 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
     end
     task.process_expired_tuple = options.process_expired_tuple or default_tuple_drop
 
+    -- validate index
+    local expire_index = box.space[space_id].index[0]
+    if options.index then
+        local index_exists = false
+        for _, index in pairs(box.space[space_id].index) do
```
olegrok, line 385: Why is `box.space[space_id].index[options.index]` not enough?
Probably I want to specify space_id, not space name - your approach doesn't allow it.

expirationd.lua
```diff
@@ -45,7 +45,11 @@ local constants = {
     -- factor for recalculation of vinyl space size
     default_vinyl_assumed_space_len_factor = 2,
     -- default function on full scan
-    default_on_full_scan = function(...) end
+    default_on_full_scan = function(...) end,
+    -- default function for start_key
+    start_key = function(...) return nil end,
```
olegrok, line 50: AFAIK `...` can't be jitted. I think we could stay arguments empty

expirationd.lua
```diff
@@ -347,6 +362,23 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
     end
     task.index = expire_index
 
+    -- check iterator_type
+    if options.iterator_type ~= nil then
+        task.iterator_type = options.iterator_type
+    end
+
+    -- check start_key
+    if options.start_key ~= nil then
```
olegrok, line 407: nit: it disallows start_key to be `box.NULL`.

And again. There are no any tests that cover your changes.

expirationd.lua
```diff
@@ -265,6 +269,16 @@ local function default_tuple_drop(space_id, args, tuple, task)
...
 
 
+-- default iterate_with function
+local function default_iterate_with(task)
+    return task.index:pairs({ task.start_key() }, { iterator = task.iterator_type })
+               :take_while(
+            function(tuple)
+                return task:process_while()
+            end
+    )
```
olegrok, line 294: Probably someone will argue me but this indents look ugly.

expirationd.lua
```diff
@@ -48,6 +48,8 @@ local constants = {
     default_on_full_scan = function(...) end,
     -- default function for start_key
     start_key = function(...) return nil end,
+    -- default function for process_while
+    process_while = function(...) return true end,
```
olegrok, line 52: The same as in https://github.com/tarantool/expirationd/pull/52/commits/ee67a1a778f012776d3da33d759a6b1888dbf0d1#r656445444

expirationd.lua
```diff
@@ -51,7 +51,9 @@ local constants = {
     -- default function for process_while
     process_while = function(...) return true end,
     -- default iterating over the loop will go in ascending index
-    iterator_type = "ALL"
+    iterator_type = "ALL",
+    -- default atomic_iteration is false, batch of items doesn't include in one transaction
+    atomic_iteration = false
```
olegrok, line 56: nit: Please add a comma at the end of line. It will help to make diff a less if you plan to add some new option. Maybe it's also related for previous commits.

expirationd.lua
```diff
@@ -100,13 +102,19 @@ local function default_do_worker_iteration(task)
     local space_len = task.vinyl_assumed_space_len
     local checked_tuples_count = 0
     local vinyl_checked_tuples_count = 0
+    if task.atomic_iteration then
```
olegrok, line 105: Will someone close transaction if task:iterate_with/expiration_process(task, tuple)/... throws an error?

expirationd.lua
```diff
@@ -608,4 +633,4 @@ return {
     get_tasks      = expirationd_show_task_list,
     run_task       = expirationd_run_task_obsolete,
     show_task_list = expirationd_show_task_list,
-}
+}
```
olegrok, line 636: Redundant change

test/unit/atomic_iteration_test.lua
```diff
@@ -0,0 +1,109 @@
...
+        table.insert(transactions, transaction)
+    end
+
+    local true_box_begin = box.begin
+
+    -- mock box.begin
+    box.begin = function ()
+        true_box_begin()
+        box.on_commit(f)
+    end
```
olegrok, line 44: It would be great to return it back at the test end.

test/unit/iterator_type_test.lua
```diff
@@ -0,0 +1,29 @@
+local expirationd = require('expirationd')
+local t = require('luatest')
+local g = t.group("iterator_type")
```
olegrok, line 3: nit: Above you use `'` but here you use `"`. I think it's better to keep the same "style" for string definition.

test/unit/process_while_test.lua
```diff
@@ -0,0 +1,47 @@
...
+
+function g.test_stopper()
+    helpers.iteration_result = {}
+    local function process_while(task)
+        if task.checked_tuples_count >= 1 then return false end
+        return true
+    end
+
+    -- for tree index spaces
+    for _, space in pairs({g.tree, g.vinyl}) do
```
olegrok, line 23: nit: I see you try to invent some kind of parametrisation for test. Try to use example from crud - https://github.com/tarantool/crud/blob/master/test/pgroup.lua .

test/unit/start_key_test.lua
```diff
@@ -0,0 +1,29 @@
+local expirationd = require('expirationd')
+local t = require('luatest')
+local g = t.group("start_key")
```
olegrok, line 3: The same as https://github.com/tarantool/expirationd/pull/52/commits/cb50f28c417266ea07c391c31f4eeeb6943e97cb#r656454532

expirationd.lua
ArtDu: You're right, it won't work. Fixed, added a test in the commit scope

expirationd.lua
ArtDu: > Probably I want to specify space_id, not space name

You probably meant options.index. Added this feature

expirationd.lua
ArtDu: Fixed for all code expirationd.lua

expirationd.lua
ArtDu: Fixed

expirationd.lua
ArtDu: I think the new version should be more like it

expirationd.lua
ArtDu: Fixed for all commits

test/unit/iterator_type_test.lua
ArtDu: Fixed for all commits and for expirationd legacy code too

expirationd.lua
```diff
@@ -322,6 +325,35 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
...
+    if options.index then
+        if type(options.index) == "string" then
+            local index_exists = false
+            for _, index in pairs(box.space[space_id].index) do
+                if index.name == options.index then
+                    expire_index = index
+                    index_exists = true
+                    break
+                end
+            end
```
olegrok, line 390:
```
box.cfg{}
_ = box.schema.space.create('test')
_ = box.space.test:create_index('pk')
_ = box.space.test:create_index('sk')

tarantool> box.space.test.index['pk'] ~= nil
---
- true
...

tarantool> box.space.test.index[1] ~= nil
---
- true
...

tarantool> box.space.test.index['sk'] ~= nil
---
- true
...

tarantool> box.space.test.index['doesnt_exist'] ~= nil
---
- false
...
```

Why do you still use a loop to check that index exists?

test/unit/custom_index_test.lua
```diff
@@ -0,0 +1,184 @@
...
+    t.assert_equals(task.index, box.space.tree.index[1])
+    task:kill()
+
+    -- index by id
+    task = expirationd.start("clean_all", g.tree.id, helpers.is_expired_true,
+            {index = 1})
+    t.assert_equals(task.index, box.space.tree.index[1])
+    task:kill()
+
+    --errors
```
olegrok, line 33: nit: `--errors` -> `-- errors`

test/helper.lua
```diff
@@ -0,0 +1,86 @@
...
+local helpers = require("luatest.helpers")
+
+t.before_suite(function()
+    t.datadir = fio.tempdir()
+    box.cfg{
+        pid_file   = t.datadir..".pid",
+        wal_dir    = t.datadir,
+        memtx_dir  = t.datadir,
+        vinyl_dir  = t.datadir,
+        listen = "127.0.0.1:3301"
```
olegrok, line 13: Do you really need it?

test/helper.lua
```diff
@@ -0,0 +1,86 @@
+local t = require("luatest")
+local fio = require("fio")
+
+local helpers = require("luatest.helpers")
+
+t.before_suite(function()
+    t.datadir = fio.tempdir()
+    box.cfg{
+        pid_file   = t.datadir..".pid",
```
olegrok, line 9: Do you really need it?

expirationd.lua
```diff
@@ -271,6 +271,9 @@ end
 --   options = {      -- (table with named options)
 --     * process_expired_tuple -- Applied to expired tuples, receives (space_id, args, tuple) as arguments;
 --                                can be nil: by default, tuples are removed.
+--     * index                 -- Name of the index to iterate on; if omitted, will use the primary index;
```
olegrok, line 309: Also it would be great to allow to pass `index_id`. It could be easily done if you fix my comment https://github.com/tarantool/expirationd/pull/52/commits/c596f8d73c3a71af2aa5139f7ca5d120eee9700d#r658927759

Just note that `create_task ` applies space_id not just space_name

test/helper.lua
```diff
@@ -0,0 +1,86 @@
+local t = require("luatest")
```
olegrok, line 1: You didn't check multikey, jsonpath and functional indexes.

expirationd.lua
```diff
@@ -256,6 +266,16 @@ local function default_tuple_drop(space_id, args, tuple)
 end
 
 
+-- default iterate_with function
+local function default_iterate_with(task)
+    return task.index:pairs(task.start_key(), { iterator = task.iterator_type })
+        :take_while(
+            function(tuple)
```
olegrok, line 288: Seems it's unused argument. Please verify your code with luacheck.

test/unit/Iterate_with_test.lua
```diff
@@ -0,0 +1,26 @@
...
+end)
+
+function g.test_passing()
+    local task = expirationd.start("clean_all", g.tree.id, helpers.is_expired_true,
+            { iterate_with = helpers.is_expired_true })
+    -- default process_while always return false, iterations never stopped by this function
+    t.assert_equals(task.iterate_with, helpers.is_expired_true)
+    task:kill()
+
+    --errors
```
olegrok, line 22: Why do you add a space between `--` and a sentence in some cases but don't do it for `--errors`?

test/unit/iterator_type_test.lua
```diff
@@ -0,0 +1,536 @@
...
+    end
+end
+
+function g.test_hash_index_eq()
+    -- without start key
+    --- throw:
+    --- main/119/worker of "clean_all" index.cc:127 E>
+    --- ER_PARTIAL_KEY: HASH index  does not support selects via a partial key (expected 1 parts, got 0).
+    --- Please Consider changing index type to TREE
+    --- Unclear error, select(nil, {iterator = "EQ"}) works with tree index and return all values, hash does not
```
olegrok, line 270: It's not unclear. Just HASH could find with "EQ" iterator only with full specified key. And such case looks really strange - "EQ" for HASH means I want to run my task to clear only single value (because hash can't non-unique)

I think such error should be thrown at validation stage.

test/unit/process_while_test.lua
```diff
@@ -0,0 +1,70 @@
...
+    local function process_while(...)
+        return false
+    end
+
+    task = expirationd.start("clean_all", g.tree.id, helpers.is_expired_true,
+            {process_while = process_while})
+    t.assert_equals(task.process_while(), false)
+    task:kill()
+
+    --errors
```
olegrok, line 30: Ditto https://github.com/tarantool/expirationd/pull/52/commits/358da8d90f4526caaf6fe840db24ca5e167564bf#r658931980

test/unit/start_key_test.lua
```diff
@@ -0,0 +1,165 @@
...
+            {start_key = 1})
+    t.assert_equals(task.start_key(), 1)
+    task:kill()
+
+    task = expirationd.start("clean_all", g.tree.id, helpers.is_expired_true,
+            { index = "multipart_index", start_key = {1, 1}})
+    t.assert_equals(task.start_key(), {1, 1})
+    task:kill()
+
+    --errors
```
olegrok, line 37: Ditto https://github.com/tarantool/expirationd/pull/52/commits/358da8d90f4526caaf6fe840db24ca5e167564bf#r658931980

expirationd.lua
olegrok: I think you can even try to reproduce it for vinyl. Expiration task delete some values in transaction another code updates the same values. Finally you transaction will be rolled back because of conflict.

Probably it's really hard to reproduce but I hope it's possible if you will give few memory for vinyl - it lead to more often yields because it will read and write directly to the disk.

test/unit/atomic_iteration_test.lua
```diff
@@ -0,0 +1,120 @@
...
+    local task = expirationd.start("clean_all", g.tree.id, helpers.is_expired_true)
+    t.assert_equals(task.atomic_iteration, false)
+    task:kill()
+
+    task = expirationd.start("clean_all", g.tree.id, helpers.is_expired_true,
+            {atomic_iteration = true})
+    t.assert_equals(task.atomic_iteration, true)
+    task:kill()
+
+    --errors
```
olegrok, line 25: Ditto https://github.com/tarantool/expirationd/pull/52/commits/358da8d90f4526caaf6fe840db24ca5e167564bf#r658931980

test/unit/atomic_iteration_test.lua
```diff
@@ -0,0 +1,120 @@
...
+                {atomic_iteration = true})
+        -- wait for tuples expired
+        helpers.retrying({}, function()
+            if space.index[0].type == "HASH" then
+                t.assert_equals(helpers.iteration_result, {{3, "1"}, {2, "2"}, {1, "3"}})
+            else
+                t.assert_equals(helpers.iteration_result, {{1, "3"}, {2, "2"}, {3, "1"}})
+            end
+        end)
+        task:kill()
```
olegrok, line 69: What happens if you try to kill a fiber during open transaction?

Exception on the first yield? It's also relevant for vinyl. I don't know about memtx to be honest. Maybe with enabled mvcc opened transaction will hang in memory (I suggest you to verify or refute my concerns).

expirationd.lua
ArtDu: Yes, I forgot to change the description

test/unit/iterator_type_test.lua
ArtDu: I wrote this to the fact that tarantool allows you to do this and gives a strange error, although as you wrote, it has to run through only one key. It seems to me that this needs to be fixed on the tarantool side, or removed from the documentation that you can use the EQ iterator for the hash index. Of course, if there are reasons not to change this in Tarantool, then I will validate this type of iterator

olegrok:
```
box.space.test:create_index('hash', {type = 'hash'})
tarantool> box.space.test.index.hash:pairs(nil, {iterator = 'EQ'})
---
- error: HASH index  does not support selects via a partial key (expected 1 parts,
    got 0). Please Consider changing index type to TREE.
...
```

It should fail on validation stage.

ArtDu: but there is no partial key here, or I'm wrong?

olegrok: `select(nil)` is the same as `select({})`. It's partial key.

expirationd.lua
ArtDu: this is a problem that without thinking I copied and pasted the idea from https://github.com/moonlibs/indexpiration, I thought it makes sense

README.md
akudiyar: I agree with a separate PR for the doc. But it is not directly related to #38, I think we need a more tight description, which options must be changed.

README.md
akudiyar: This option description needs rewriting

test/unit/iterator_type_test.lua
akudiyar: Let's add a validation for `nil` index value.

.github/workflows/fast_testing.yaml
akudiyar: Let's do this in a separate task. Completion of this feature is urgent for the customer.

test/unit/iterator_type_test.lua
olegrok: > Let's add validation for `nil` index value.

It doesn't work so. You also should ban `{}` and `{box.NULL}`. It's bad way. Just check with `box.space.test.index.hash:pairs(nil, {iterator = 'EQ'})`.

expirationd.lua
ArtDu: I think these tests should be in the tarantool / tarantool repository, in my opinion, this is not entirely relevant for this task

test/unit/atomic_iteration_test.lua
ArtDu: There should also be tarantool/tarantool tests for this

olegrok: No. It's your responsibility to close transaction.

olegrok: I mean that it would be great to introduce some flag that fiber will check on each iteration. E.g. is_finished. After "is_finished" is changed to true fiber should gracefully shutdown its work.

expirationd.lua
olegrok: No. It's responsibility of user code to close transactions. Your code could turn Tarantool in inconsistent condition. Please fix it and handle all errors properly.

olegrok: After transaction throws an error fiber will die.

Should user manually restart it's task after your unsafe code terminate this task? Even task will be alive you can't open transaction again. Because box.begin() could be called twice.

```
tarantool> box.begin() box.begin()
---
- error: 'Operation is not permitted when there is an active transaction '
...
```

Do not justify the unsafety of your code by saying that Tarantool should fix all this for you.

test/unit/atomic_iteration_test.lua
```diff
@@ -0,0 +1,165 @@
...
+    -- wait for tuples expired
+    fiber.sleep(3)
+
+    task:kill()
+    for i = 1,10 do
+        fiber.kill(updaters[i]:id())
+    end
+
+    t.assert(box.stat.vinyl().tx.conflict > 0)
+    t.assert_equals(box.stat.vinyl().tx.conflict, box.stat.vinyl().tx.rollback)
```
olegrok, line 151: And seems we also should check that task still works and after there is no load anymore tuples are expired.

test/unit/atomic_iteration_test.lua
```diff
@@ -0,0 +1,165 @@
...
+
+    local task = expirationd.start("clean_all", g.vinyl.id, helpers.is_expired_debug,
+            {atomic_iteration = true})
+
+    -- wait for tuples expired
+    fiber.sleep(3)
+
+    task:kill()
+    for i = 1,10 do
+        fiber.kill(updaters[i]:id())
```
olegrok, line 147: Why not `updaters[i]:cancel()`?

test/unit/custom_index_test.lua
```diff
@@ -0,0 +1,257 @@
...
+                {1, "1", nil, nil, nil, nil, {data = {{name = 'A'},
+                                                      {name = 'B'}},
+                                              extra_field = 1}}
+            })
+        end)
+        task:kill()
+    end
+end
+
+function g.test_memtx_tree_functional_index()
```
olegrok, line 181: Tests for json/functional/multikey indexes should be disabled for 1.10 at least

test/unit/atomic_iteration_test.lua
olegrok: And probably let's check `box.stat.vinyl().tx.transactions == 0` - there are no opened transactions.

test/unit/atomic_iteration_test.lua
```diff
@@ -0,0 +1,165 @@
...
+    end
+
+    -- unmock
+    box.begin = true_box_begin
+end
+
+function g.test_mvcc_vinyl_tx_conflict()
+    for i = 1,10 do
+        g.vinyl:insert({i, tostring(i), nil, nil, 0})
+    end
```
olegrok, line 126: Probably it would be great to add comment that we don't check Tarantool or vinyl as engine. We check that expiration task continue work even in case of conflict.

expirationd.lua
```diff
@@ -215,14 +206,18 @@ local Task_methods = {
...
+            while self.guardian_fiber:status() ~= "dead" do
                 fiber.sleep(0.01)
             end
             self.guardian_fiber = nil
         end
         if (get_fiber_id(self.worker_fiber) ~= 0) then
-            self.worker_fiber:cancel()
-            while self.worker_fiber:status() ~= 'dead' do
+            if self.atomic_iteration then
+                self.worker_canceled = true
```
olegrok, line 216: I suggest to do it even atomic_iteration is force

expirationd.lua
```diff
@@ -215,14 +206,18 @@ local Task_methods = {
...
+            else
+                self.worker_fiber:cancel()
+            end
+            while self.worker_fiber:status() ~= "dead" do
                 fiber.sleep(0.01)
             end
```
olegrok, line 222: Why not `self.worker_fiber:set_joinable(true)` + `self.worker_fiber:join()`?

test/unit/Iterate_with_test.lua
ArtDu: Copy paste problem

test/unit/iterator_type_test.lua
ArtDu: There was such a check, I just finished the test to check this validation

test/unit/new_features_test.lua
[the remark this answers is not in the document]
ArtDu: We decided that this is normal behavior since nil and {} are partial keys 
https://github.com/tarantool/expirationd/pull/52#discussion_r658933380

README.md
ArtDu: deleted unnecessary words

README.md
ArtDu: Now both(luatest, taptest) and those tests are launched in the make

expirationd.lua
ArtDu: After personal messaging, we understood that the error will fall into pcall and the fiber will die and a new one will start. For exact rollback of the transaction, box.rollback() was added before the fiber dies

Worker loop:
```lua
...
local state, err = pcall(task.do_worker_iteration, task)
if state then
    task.on_full_scan_success(task)
else
    task.on_full_scan_error(task, err)
end

task.on_full_scan_complete(task)
if not state then
    box.rollback()
    error(err)
end
...
```

```lua
local function get_fiber_id(fiber)
    local fid = 0
    if fiber ~= nil and fiber:status() ~= "dead" then
        fid = fiber:id()
    end
    return fid
end

local function guardian_loop(task)
    -- detach the guardian from the creator and attach it to sched
    fiber.name(string.format("guardian of %q", task.name), { truncate = true })

    while true do
        -- if fiber doesn't exist
        if get_fiber_id(task.worker_fiber) == 0 then
            -- create worker fiber
            task.worker_fiber = fiber.create(worker_loop, task)

            log.info("expiration: task %q restarted", task.name)
            task.restarts = task.restarts + 1
        end
        fiber.sleep(constants.check_interval)
    end
end
```

test/unit/atomic_iteration_test.lua
ArtDu: Was added worker_canceled flag for that

test/unit/atomic_iteration_test.lua
ArtDu: Fixed

test/unit/atomic_iteration_test.lua
ArtDu: fixed

README.md
ArtDu: Ticket: https://github.com/tarantool/expirationd/issues/60

test/unit/process_while_test.lua
ArtDu: since this is a nit, I decided to leave it as it is, since it is important to close this task faster

expirationd.lua
ArtDu: Probably you meant false, I left it as it was, because it is faster, and for the transaction it is more important that it completes correctly. Maybe it needs to be discussed

expirationd.lua
ArtDu: I get a weird vinyl and join bug. In some tests data is not deleted here:
```lua
-- default process_expired_tuple function
local function default_tuple_drop(space_id, args, tuple)
    box.space[space_id]:delete(construct_key(space_id, tuple))
end
```

But if we write some sleep, everything work fine:
```lua
-- default process_expired_tuple function
local function default_tuple_drop(space_id, args, tuple)
    box.space[space_id]:delete(construct_key(space_id, tuple))
    fiber.sleep(0.1)
end
```

olegrok: How is it related to my comment?

```
while self.worker_fiber:status() ~= "dead" do
                fiber.sleep(0.01)
            end
```

Here you wait until fiber is terminated. I suggest to do it without a loop and use fiber.join.

ArtDu: Yes, I say that I changed loop to fiber.join and got this bug https://github.com/tarantool/expirationd/pull/52#discussion_r662213292

olegrok: It's quite strange. Could it be workarounded for at test level?

ArtDu: Insert custom deleter, but this is wierd, can we just leave loop with sleep?

olegrok: Anyway you put such code without any comment and how should people in future understand it?

ArtDu: I have helpers.retry to check the result, sleep does not work in the test, everything breaks in the section with deletion. What code have I put without comments and that it is not clear? sleep with a status check for waiting for a fiber was in the original implementation. Perhaps somewhere additional comments are needed, but I would like to know which place you specifically point to

olegrok: Before I thought that this code really simply wait when fiber will be finished. But as I understand you also state that this code makes vinyl tests work. It's strange, not obvious.

olegrok: Also with
```patch
index 34367b6..0938702 100644
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -205,21 +205,19 @@ local Task_methods = {
     end,
     stop = function (self)
         if (get_fiber_id(self.guardian_fiber) ~= 0) then
+            self.guardian_fiber:set_joinable(true)
             self.guardian_fiber:cancel()
-            while self.guardian_fiber:status() ~= "dead" do
-                fiber.sleep(0.01)
-            end
+            self.guardian_fiber:join()
             self.guardian_fiber = nil
         end
         if (get_fiber_id(self.worker_fiber) ~= 0) then
+            self.worker_fiber:set_joinable(true)
             if self.atomic_iteration then
                 self.worker_canceled = true
             else
                 self.worker_fiber:cancel()
             end
-            while self.worker_fiber:status() ~= "dead" do
-                fiber.sleep(0.01)
-            end
+            self.worker_fiber:join()
             self.worker_fiber = nil
         end
     end,
```

I have:
```bash
➜  expirationd git:(indexpiration) ✗ .rocks/bin/luatest -v
Started on Thu Jul  1 16:43:19 2021
    atomic_iteration.test_passing ... (1.005s) Ok
    atomic_iteration.test_memtx ... (4.018s) Ok
    atomic_iteration.test_mvcc_vinyl_tx_conflict ... (4.071s) Ok
    atomic_iteration.test_kill_task ... (0.275s) Ok
    custom_index.test_passing ... (0.001s) Ok
    custom_index.test_tree_index ... (0.413s) Ok
    custom_index.test_tree_index_non_unique ... (0.208s) Ok
    custom_index.test_tree_index_multipart ... (0.212s) Ok
    custom_index.test_tree_index_json_path ... (0.210s) Ok
    custom_index.test_tree_index_multikey ... (2.009s) Ok
    custom_index.test_memtx_tree_functional_index ... (1.037s) Ok
    custom_index.test_hash_index ... (0.210s) Ok
    custom_index.test_hash_index_multipart ... (0.107s) Ok
    iterate_with.test_passing ... (0.002s) Ok
    iterator_type.test_passing_errors ... (0.001s) Ok
    iterator_type.test_passing_all ... (0.002s) Ok
    iterator_type.test_passing_eq ... (0.001s) Ok
    iterator_type.test_passing_gt ... (0.001s) Ok
    iterator_type.test_passing_req ... (0.001s) Ok
    iterator_type.test_passing_ge ... (0.001s) Ok
    iterator_type.test_passing_lt ... (0.001s) Ok
    iterator_type.test_passing_le ... (0.001s) Ok
    iterator_type.test_tree_index_all ... (0.413s) Ok
    iterator_type.test_hash_index_all ... (0.208s) Ok
    iterator_type.test_tree_index_eq ... (0.413s) Ok
    iterator_type.test_hash_index_eq ... (1.007s) Ok
    iterator_type.test_tree_index_gt ... (2.213s) Ok
    iterator_type.test_hash_index_gt ... (1.130s) Ok
    iterator_type.test_tree_index_req ... (0.409s) Ok
    iterator_type.test_tree_index_ge ... (0.409s) Ok
    iterator_type.test_tree_index_lt ... (2.216s) Ok
    iterator_type.test_tree_index_le ... (0.419s) Ok
    process_while.test_passing ... (0.003s) Ok
    process_while.test_tree_index ... (2.007s) Ok
    process_while.test_hash_index ... (1.006s) Ok
    start_key.test_passing ... (0.002s) Ok
    start_key.test_tree_index ... (0.622s) Ok
    start_key.test_hash_index ... (0.308s) Ok
=========================================================
Ran 38 tests in 26.612 seconds, 38 successes, 0 failures
```

It seems to be ok.
```
Tarantool Enterprise 2.8.1-0-ge2a1ec0c2-r407-macos
Target: Darwin-x86_64-RelWithDebInfo
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/opt/actions-runners/sdk_01/_work/sdk/sdk/build.sdk/tarantool-2.8/static-build/tarantool-prefix -DENABLE_BACKTRACE=TRUE
Compiler: /Library/Developer/CommandLineTools/usr/bin/cc /Library/Developer/CommandLineTools/usr/bin/c++
C_FLAGS: -Wno-unknown-pragmas -fexceptions -funwind-tables -fno-omit-frame-pointer -fno-stack-protector -fno-common -msse2 -std=c11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-gnu-alignof-expression
CXX_FLAGS: -Wno-unknown-pragmas -fexceptions -funwind-tables -fno-omit-frame-pointer -fno-stack-protector -fno-common -msse2 -std=c++11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-invalid-offsetof -Wno-gnu-alignof-expression
```

ArtDu: I did as indicated in the documentation, maybe I have an old version of the tarantula and some bug is not fixed. But writing set_joinable like you made it work. But this doesn't work

```patch
index 34367b6..bfc3f84 100644
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -179,7 +179,8 @@ local function guardian_loop(task)
         -- if fiber doesn't exist
         if get_fiber_id(task.worker_fiber) == 0 then
             -- create worker fiber
-            task.worker_fiber = fiber.create(worker_loop, task)
+            task.worker_fiber = fiber.new(worker_loop, task)
+            task.worker_fiber:set_joinable(true)
 
             log.info("expiration: task %q restarted", task.name)
             task.restarts = task.restarts + 1
@@ -201,14 +202,13 @@ end
 local Task_methods = {
     start = function (self)
         self:stop()
-        self.guardian_fiber = fiber.create(guardian_loop, self)
+        self.guardian_fiber = fiber.new(guardian_loop, self)
+        self.guardian_fiber:set_joinable(true)
     end,
     stop = function (self)
         if (get_fiber_id(self.guardian_fiber) ~= 0) then
             self.guardian_fiber:cancel()
-            while self.guardian_fiber:status() ~= "dead" do
-                fiber.sleep(0.01)
-            end
+            self.guardian_fiber:join()
             self.guardian_fiber = nil
         end
         if (get_fiber_id(self.worker_fiber) ~= 0) then
@@ -217,9 +217,7 @@ local Task_methods = {
             else
                 self.worker_fiber:cancel()
             end
-            while self.worker_fiber:status() ~= "dead" do
-                fiber.sleep(0.01)
-            end
+            self.worker_fiber:join()
             self.worker_fiber = nil
         end
     end,
```
 TESTS:
```
Started on Thu Jul  1 16:51:50 2021
    atomic_iteration.test_passing ... (2.010s) Ok
    atomic_iteration.test_memtx ... (6.021s) Ok
    atomic_iteration.test_mvcc_vinyl_tx_conflict ... (8.259s) fail
...arantool/expirationd/test/unit/atomic_iteration_test.lua:153: expected:
{}
actual:
{
    [1, '1', null, null, 33],
    [2, '2', null, null, 33],
    [3, '3', null, null, 33],
    [4, '4', null, null, 33],
    [5, '5', null, null, 33],
    [6, '6', null, null, 33],
    [7, '7', null, null, 33],
    [8, '8', null, null, 32],
    [9, '9', null, null, 32],
    [10, '10', null, null, 33],
}
    atomic_iteration.test_kill_task ... (1.574s) Ok
    custom_index.test_passing ... (3.009s) Ok
    custom_index.test_tree_index ... (0.412s) Ok
    custom_index.test_tree_index_non_unique ... (0.205s) Ok
    custom_index.test_tree_index_multipart ... (0.206s) Ok
    custom_index.test_tree_index_json_path ... (0.206s) Ok
    custom_index.test_tree_index_multikey ... (5.097s) fail
...in/tarantool/expirationd/test/unit/custom_index_test.lua:172: expected:
{{1, "1", [7] = {data = {{name = "A"}, {name = "B"}}, extra_field = 1}}}
actual:
{
    [1, '1', null, null, null, null, {'data': [{'name': 'A'}, {'name': 'B'}], 'extra_field': 1}],
    [1, '1', null, null, null, null, {'data': [{'name': 'A'}, {'name': 'B'}], 'extra_field': 1}],
}
    custom_index.test_memtx_tree_functional_index ... (0.105s) Ok
    custom_index.test_hash_index ... (0.209s) Ok
    custom_index.test_hash_index_multipart ... (0.104s) Ok
    iterate_with.test_passing ... (1.007s) Ok
    iterator_type.test_passing_errors ... (0.001s) Ok
    iterator_type.test_passing_all ... (4.009s) Ok
    iterator_type.test_passing_eq ... (2.011s) Ok
    iterator_type.test_passing_gt ... (2.006s) Ok
    iterator_type.test_passing_req ... (1.002s) Ok
    iterator_type.test_passing_ge ... (1.003s) Ok
    iterator_type.test_passing_lt ... (1.006s) Ok
    iterator_type.test_passing_le ... (1.005s) Ok
    iterator_type.test_tree_index_all ... (0.410s) Ok
    iterator_type.test_hash_index_all ... (0.205s) Ok
    iterator_type.test_tree_index_eq ... (0.408s) Ok
    iterator_type.test_hash_index_eq ... (0.104s) Ok
    iterator_type.test_tree_index_gt ... (0.412s) Ok
    iterator_type.test_hash_index_gt ... (1.137s) Ok
    iterator_type.test_tree_index_req ... (0.415s) Ok
    iterator_type.test_tree_index_ge ... (0.411s) Ok
    iterator_type.test_tree_index_lt ... (0.408s) Ok
    iterator_type.test_tree_index_le ... (0.408s) Ok
    process_while.test_passing ... (2.008s) Ok
    process_while.test_tree_index ... (0.207s) Ok
    process_while.test_hash_index ... (0.103s) Ok
    start_key.test_passing ... (4.011s) Ok
    start_key.test_tree_index ... (0.621s) Ok
    start_key.test_hash_index ... (0.314s) Ok
=========================================================
```

TARANTOOL:
```
Tarantool Enterprise 2.5.3-0-gf93e48013-r393-macos
Target: Darwin-x86_64-RelWithDebInfo
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/opt/actions-runners/sdk_01/_work/sdk/sdk/build.sdk/tarantool-2.5/static-build/tarantool-prefix -DENABLE_BACKTRACE=TRUE
Compiler: /Library/Developer/CommandLineTools/usr/bin/cc /Library/Developer/CommandLineTools/usr/bin/c++
C_FLAGS: -Wno-unknown-pragmas -fexceptions -funwind-tables -fno-omit-frame-pointer -fno-stack-protector -fno-common -msse2 -std=c11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-gnu-alignof-expression
CXX_FLAGS: -Wno-unknown-pragmas -fexceptions -funwind-tables -fno-omit-frame-pointer -fno-stack-protector -fno-common -msse2 -std=c++11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-invalid-offsetof -Wno-gnu-alignof-expression
```

ArtDu: I think the problem is in fiber.new, changing back to create everything worked for my above

olegrok: Well... Ok. Let's stay as is.

.github/workflows/fast_testing.yaml
ArtDu: https://github.com/tarantool/expirationd/issues/61

akudiyar (approved):

olegrok (approved):
I give up. Probably we should start to file issues for some comments. Otherwise there is a risk that this patch never will be merged.

.github/workflows/fast_testing.yaml
ligurio: Discussed and decided to postpone it and do it in the scope of another ticket.

README.md
ligurio: Why do you need to change the formatting for options not related to your changes?

Makefile
```diff
@@ -3,6 +3,7 @@ all:
 
 .PHONY: test
 test:
+	.rocks/bin/luatest -v
```
ligurio, line 6: This commit contains changes for README. Change in Makefile should be moved to commit 
"ci: installation, caching and running luatest".

ligurio (changes requested):
- Limit commit message width in commits "Iterate_with and process_while" and "Start key and iterator type"
- We decided to move some requested on review changes to the backlog. Please make sure all appropriated tickets submitted.
  - JSON-path index error https://github.com/tarantool/expirationd/issues/59
  - Unclear options https://github.com/tarantool/expirationd/issues/60
  - Convert TAP-style tests to luatest-style test https://github.com/tarantool/expirationd/issues/61
- We need to enable new tests as soon as you added them. So put luatest related changes to commit
where first test is appeared (it is "Iterate over any index").
- Changes are not atomic now. I see commits with new functionality without any documentation in a README.
If we don't need to document these changes then please explain why.
- Why do you mix double quotes and single quotes in a whole patchset?
- Latest commit should contain a "Closes #50", otherwise it is not clear finished task or not.

Makefile
ArtDu: Moved

ArtDu:
> * Limit commit message width in commits "Iterate_with and process_while" and "Start key and iterator type"
> * We decided to move some requested on review changes to the backlog. Please make sure all appropriated tickets submitted.
>   
>   * Unclear options [Unclear options #60](https://github.com/tarantool/expirationd/issues/60)
>   * Convert TAP-style tests to luatest-style test [Rewrite tests from tap to luatest #61](https://github.com/tarantool/expirationd/issues/61)
> * We need to enable new tests as soon as you added them. So put luatest related changes to commit
>   where first test is appeared (it is "Iterate over any index").
> * Changes are not atomic now. I see commits with new functionality without any documentation in a README.
>   If we don't need to document these changes then please explain why.
> * Why do you mix double quotes and single quotes in a whole patchset?
> * Latest commit should contain a "Closes [RFC: Adding Iterations over Secondary Indexes  #50](https://github.com/tarantool/expirationd/issues/50)", otherwise it is not clear finished task or not.

* Reduced the body size of all commits to 72 characters.
* Luatest is added to ci with the first appearance of tests ("Iterate over any index" commit).
* Changes are atomic now. Each commit on changing the functionality contains the addition of functionality, its description in the readme and testing it in the luatest.
* Mix double quotes and single quotes were not in the code. The outer ones are double quotes, the same style has been added to the README for greater consistency.
* Added closes: #50 in last commit body message.

README.md
```diff
@@ -31,35 +31,30 @@ expirationd.start(job_name, space.id, is_expired, {
...
+    * `process_expired_tuple` - Applied to expired tuples, receives (space_id, args, tuple) as arguments.
+     Can be nil: by default, tuples are removed.
+    * `tuples_per_iteration` - Number of tuples to check in one batch (iteration). Default is 1024.
+    * `on_full_scan_start` - Function to call before starting a tuple scan.
+    * `on_full_scan_complete` - Function to call after completing a full scan.
+    * `on_full_scan_success` - Function to call after successfully completing a full scan.
+    * `on_full_scan_error` - Function to call after terminating a full scan due to an error.
+    * `args` - Passed to is_tuple_expired and process_expired_tuple() as an additional context.
+    * `full_scan_time` - Time required for a full index scan (in seconds).
+    * `iteration_delay` - Max sleep time between batchs (in seconds).
```
ligurio, line 105: batches

expirationd.lua
```diff
@@ -71,13 +73,15 @@ local function expiration_process(task, tuple)
...
 
+-- yield for some time
 local function suspend_basic(scan_space, task, len)
     local delay = (task.tuples_per_iteration * task.full_scan_time)
     delay = math.min(delay / len, task.iteration_delay)
     fiber.sleep(delay)
 end
 
 local function suspend(scan_space, task)
+    -- Return the number of tuples in the space
```
ligurio, line 92: Why previous added comments are above function and this one below function?

expirationd.lua
```diff
@@ -298,27 +303,28 @@ end
...
+--     * on_full_scan_complete -- Function to call after completing a full scan iteration.
+--     * on_full_scan_success  -- Function to call after successfully completing a full scan iteration.
+--     * on_full_scan_error    -- Function to call after terminating a full scan due to an error.
+--     * args                  -- Passed to is_tuple_expired and
+--                                process_expired_tuple() as additional context.
+--     * tuples_per_iteration  -- Number of tuples to check in one batch (iteration); default is 1024.
+--     * full_scan_time        -- Time required for a full index scan (in seconds).
+--     * iteration_delay       -- Max sleep time between iterations (in seconds).
+--     * full_scan_delay       -- Sleep time between full scans (in seconds).
+--     * force                 -- Run task even on replica.
```
ligurio: How change with formatting is related to this commit?

.github/workflows/fast_testing.yaml
```diff
@@ -35,4 +35,17 @@ jobs:
...
+        uses: actions/cache@v2
+        id: cache-rocks
+        with:
+          path: .rocks/
+          key: cache-rocks-${{ matrix.tarantool }}-01
+
+      - name: Install requirements
+        run: |
+          tarantoolctl rocks install luatest 0.5.0
+        if: steps.cache-rocks.outputs.cache-hit != 'true'
```
ligurio, line 48: You are enabled luatest in CI, but luatest is unused. You add luatest-style tests later in further commits.

Makefile
ligurio: The same: no luatest-style tests in a project. You add luatest-style tests later in further commits.

ligurio (changes requested):

expirationd.lua
ArtDu: Because this comment is not for a function, but for the line below

README.md
ArtDu: Fixed

.github/workflows/fast_testing.yaml
ArtDu: Fixed. Luatest appears in ci along with the first luatest's tests

expirationd.lua
ArtDu: commit msg: `Comments, readme, and responses to errors are presented in a more uniform form.`

ligurio (approved):

PR #55, opened 2021-05-28
ci: add initial CI

ArtDu:
Added source rock and rockspec at release deployment. See https://github.com/tarantool/smtp/pull/41.
As testing, tests are run on both luatest and tap test.

travis-ci: Removing travis as a build and deployment tool. Deployment of deb \ rpm packages does not work temporarily. Added a ticket for the future implementation of this functionality on gh actions (https://github.com/tarantool/expirationd/issues/56).

It is also worth adding luacheck linter in the future (https://github.com/tarantool/expirationd/issues/57)


akudiyar (approved):

.github/workflows/publish.yml
```diff
@@ -0,0 +1,62 @@
...
+      # `tarantoolctl rocks pack <rockspec>` creates a source
+      # tarball. It speeds up
+      # `tarantoolctl rocks install <module_name> <version>` and
+      # frees it from dependency on git.
+      #
+      # Important: Don't confuse this command with
+      # `tarantoolctl rocks pack <module_name> [<version>]`, which
+      # creates a **binary** rock or .all.rock (see [1]). Don't
+      # upload a binary rock of a Lua/C module to
+      # rocks.tarantool.org. Lua/C modules are platform dependent.
```
olegrok, line 43: But it's not so for expirationd because it's written in pure Lua. 
In current case it's possible to upload .all.rock (as e.g. in [errors](https://github.com/tarantool/errors/blob/master/.github/workflows/publish.yml#L46)).

ArtDu: Stole from the error repository. Also create tagged rockspec via `tarantoolctl rocks new_version --tag ${{ env.TAG }}` instead of sed code.

.github/workflows/testing.yaml
```diff
@@ -0,0 +1,58 @@
...
+        uses: actions/cache@v2
+        id: cache-rocks
+        with:
+          path: .rocks/
+          key: cache-rocks-${{ matrix.tarantool }}-01
+
+      - name: Install requirements
+        run: |
+          tarantoolctl rocks install luatest 0.5.0
+          CMAKE_DUMMY_WEBUI=YES tarantoolctl rocks install cartridge 2.5.1
```
olegrok, line 48: Personally, I'm not sure that it's really needed for now.
The final decision will be made by @Totktonada.

ArtDu: In principle, yes, this can be added when really needed for hot-reload tests.

olegrok:
Also please remove trailing dots from commit message titles.

ArtDu:
> Also please remove trailing dots from commit message titles.
Was done.

olegrok (approved):
I don't know that we should do with https://github.com/tarantool/expirationd/pull/55#discussion_r643225828 .
But patch seems to be ok.

.github/workflows/publish.yml
Totktonada: The `new_version` produces working, but non quite readable result. I would prefer `sed`. See details [here](https://github.com/tarantool/smtp/pull/38#discussion_r621460012).

.github/workflows/testing.yaml
Totktonada: Sure, it would be better to add luatest installation and so on into CI, when we'll have something to run using luatest.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,61 @@
...
+          tarantool-version: '2.5'
+
+      # Create a rockspec for the release.
+      - run: tarantoolctl rocks new_version --tag ${{ env.TAG }}
+
+      # Create a .all.rock tarball.
+      - run: tarantoolctl rocks install expirationd-${{ env.TAG }}-1.rockspec
+      - run: tarantoolctl rocks pack expirationd ${{ env.TAG }}
+
+      # Upload .rockspec and .src.rock.
```
Totktonada, line 55: We upload .rockspec and .all.rock here: the comment is misleading.

BTW, is there any sense to upload .src.rock as well, when we upload .all.rock? (I don't know.)

Makefile
```diff
@@ -1,6 +1,7 @@
 all:
 	@echo "Only tests are available: make test"
 
+.PHONY: test
```
Totktonada, line 4: It is relevant to the commit, where you added luatest based tests within the `test/` directory. Not here.

Totktonada:
We agreed with @ArtDu and @akudiyar that I'll fix remaining comments and push the PR.

Hey! Just observed: you removed packaging at all, not just from CI. I'll revert it and just drop integration with Travis CI.

.github/workflows/publish.yml
Totktonada: > BTW, is there any sense to upload .src.rock as well, when we upload .all.rock? (I don't know.)

Asked @rosik and @olegrok: they don't know cases, when it may be needed.

.github/workflows/publish.yml
Totktonada: I rewritten this comment (left causion re binary rocks anyway). I use `sed` for generating a release rockspec. So, we can use tarantool 1.10 and avoid usage of unsupported tarantool release. The resulting release rockspec looks better now, but when copy-pasting the `sed` code for other modules, we should pay attention to whether single or double quotes are used in the scm-1 rockspec. Not a silver bullet...

.github/workflows/testing.yaml
Totktonada: I removed this code hunk.

.github/workflows/publish.yml
Totktonada: Fixed the comment.

Makefile
Totktonada: Removed from the patch.

Totktonada (approved):

Totktonada:
Updated patches according to my comments above. Added more details to commit messages.

Issue #61, opened 2021-07-01
Rewrite tests from tap to luatest


ArtDu:
https://github.com/tarantool/expirationd/pull/52#discussion_r654379246

ligurio:
TAP tests require more than one parameter
```
test:
        .rocks/bin/luatest -v
        rm -rf ${CLEANUP_FILES}
        INDEX_TYPE='TREE' SPACE_TYPE='vinyl' ./test.lua
        rm -rf ${CLEANUP_FILES}
        INDEX_TYPE='HASH' ./test.lua
        rm -rf ${CLEANUP_FILES}
        INDEX_TYPE='TREE' ./test.lua
        rm -rf ${CLEANUP_FILES}
```
but parameterization with more than one parameter is in progress feature -
https://github.com/tarantool/luatest/pull/163

ligurio:
luatest 0.5.5 includes parameterize support
https://github.com/tarantool/luatest/commit/717cf1d03440f1134fbd8010775084878c1c3949

ligurio:
> Actually there are two blockers: #81 and #80.

Both tickets closed.

ArtDu:
> > Actually there are two blockers: #81 and #80.
> 
> Both tickets closed.
Ok then I start rewriting

ArtDu:
@ligurio, what do you think is better to rewrite the tests as close to the original as possible? or use already existing spaces to test the main essence of taptests? 

Do I need to make the same logs as they were?
```lua
2021-11-23 19:27:45.310 [13935] main/103/test.lua I> ------------- print -----------
2021-11-23 19:27:45.310 [13935] main/103/test.lua I> before print space "origin"
2021-11-23 19:27:45.310 [13935] main/103/test.lua I> [0, 'test_0@tarantool.org', 1637684865]
2021-11-23 19:27:45.310 [13935] main/103/test.lua I> [1, 'test_1@tarantool.org', 1637684866]
2021-11-23 19:27:45.310 [13935] main/103/test.lua I> [2, 'test_2@tarantool.org', 1637684867]
2021-11-23 19:27:45.310 [13935] main/103/test.lua I> [3, 'test_3@tarantool.org', 1637684868]
2021-11-23 19:27:45.310 [13935] main/103/test.lua I> [4, 'test_4@tarantool.org', 1637684869]
2021-11-23 19:27:45.310 [13935] main/103/test.lua I> [5, 'test_5@tarantool.org', 1637684870]
[...]

PR #62, opened 2021-07-06
Enable Lua source code analysis with luacheck

### LINKED ISSUE #57: Add linter(luacheck) into ci

ligurio:
Added luacheck config and fixed a number of luacheck warnings.
Tarantool's Lua style guide allows to ignore some luacheck warnings [1].
File with TAP tests added to ignore list because we plan to rewrite
these tests to luatest style, see [2].

1. https://www.tarantool.io/en/doc/latest/dev_guide/lua_style_guide/#luacheck
2. https://github.com/tarantool/expirationd/issues/61

Closes #57

### DIFF
--- a/.github/workflows/fast_testing.yaml
+++ b/.github/workflows/fast_testing.yaml
@@ -40,11 +40,14 @@ jobs:
         id: cache-rocks
         with:
           path: .rocks/
-          key: cache-rocks-${{ matrix.tarantool }}-01
+          key: cache-rocks-${{ matrix.tarantool }}-02
 
       - name: Install requirements
         run: |
           tarantoolctl rocks install luatest 0.5.0
+          tarantoolctl rocks install luacheck 0.26.0
         if: steps.cache-rocks.outputs.cache-hit != 'true'
+
+      - run: make check
 
       - run: make test
--- a/test/unit/atomic_iteration_test.lua
+++ b/test/unit/atomic_iteration_test.lua
@@ -36,7 +36,7 @@ function g.test_memtx()
     local function f(iterator)
         local transaction = {}
         -- old / new_tuple is not passed for vinyl
-        for request_number, old_tuple, new_tuple, space_id in iterator() do
+        for _, old_tuple in iterator() do
             table.insert(transaction, old_tuple:totable())
         end
         table.insert(transactions, transaction)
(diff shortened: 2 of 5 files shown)


test/unit/atomic_iteration_test.lua
```diff
@@ -36,7 +36,7 @@ function g.test_memtx()
     local function f(iterator)
         local transaction = {}
         -- old / new_tuple is not passed for vinyl
-        for request_number, old_tuple, new_tuple, space_id in iterator() do
+        for _, old_tuple, _, _ in iterator() do
```
LeonidVas, line 39: Why not just `for _, old_tuple in iterator() do`?

LeonidVas (approved):
Generally LGTM.
See several small questions bellow.

test/unit/atomic_iteration_test.lua
ligurio: When we replace var name with `_` we lost some information about items for the source code reader.
But I don't want to keep it as is and suppress luacheck warning inline. So I replaced variables to `_` but keep all four variables to notice for the reader that we have 4 variables on each loop.

.github/workflows/fast_testing.yaml
```diff
@@ -40,11 +40,14 @@ jobs:
         id: cache-rocks
         with:
           path: .rocks/
-          key: cache-rocks-${{ matrix.tarantool }}-01
+          key: cache-rocks-${{ matrix.tarantool }}-02
```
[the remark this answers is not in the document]
ligurio: GH Actions don't install new rock without bumping postfix in cache key, I believe the reason is that it uses an old key for cache where new rock is absent.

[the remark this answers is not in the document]
ligurio: In short, you need to change the cache path every time when you update a list of installed rocks.

test/unit/atomic_iteration_test.lua
LeonidVas: As for me, for saving information, in this case, the comment looks preferable, because now it seems like half of the developerfriendly behavior. "I will tell you the number of the elements, but I will not tell what those elements are".
This is just my opinion and I don't object if this piece of code stays "as is".
Approve was set earlier.

ligurio: Well, I keep only `_, old_tuple`.
Thanks for the review!

Issue #64, opened 2021-07-22
Infinity increase of fiber counter


ArtDu:
If we drop a space that has an expirationd task, then it will endlessly create fibers. I am not claiming that it will be a fiber bomb, since fibers die immediately due to the lack of space error. But the fiber counter is increasing exactly

![image](https://user-images.githubusercontent.com/22258443/126665699-f617807f-1195-4ef1-a9b4-6551e3c3c337.png)

The user must control the existence of the space and the expirationd tasks himself, but the code is best made as safe as possible

PR #65, opened 2021-07-22
test: space manipulations

### LINKED ISSUE #63: Add test for space:rename() when expirationd is running on rename space
What if expirationd will be runed on some space and that space will be renamed by space:rename? Seems that this case must be tested and fixed if there any misbehavior

ArtDu:
Checking how expirationd will work in unforeseen situations.
Rename/truncate/drop space.

Closes #63

### DIFF
--- a/test/unit/space_manipulation_test.lua
+++ b/test/unit/space_manipulation_test.lua
@@ -1,0 +1,107 @@
+local expirationd = require("expirationd")
+local t = require("luatest")
+local g = t.group("space_manipulation")
+
+local helpers = require("test.helper")
+
+g.before_all(function()
+    helpers.init_spaces(g)
+end)
+
+g.after_each(function()
+    helpers.truncate_spaces(g)
+end)
+
+function g.test_rename()
+    helpers.iteration_result = {}
+
+    g.tree:insert({1, "3"})
+    g.tree:insert({2, "2"})
+    g.tree:insert({3, "1"})
+
+    local task = expirationd.start("clean_all", g.tree.id, helpers.is_expired_debug)
+    -- wait for tuples expired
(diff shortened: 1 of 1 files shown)


vrogach2020:
 - we must declare expected expirationD behavior and check it in test instead of just smoke-testing.

ligurio:
Also it should close the ticket https://github.com/tarantool/expirationd/issues/64

ArtDu:
We should also check the drop of the index

oleg-jukovec:
Expected expirationd behavior should become clearer after resolving #68  and #116 (create, drop) . Also it may be related to #112 (rename case).

I will return to the pull request after we resolve these issues.

oleg-jukovec (commented):
An expected behavior of expirationd should become clearer after resolving https://github.com/tarantool/expirationd/issues/68 and https://github.com/tarantool/expirationd/issues/116 (create, drop) . Also it may be related to https://github.com/tarantool/expirationd/issues/112 (rename) , https://github.com/tarantool/expirationd/issues/64 (rename, drop).

We should return to the pull request after we resolve these issues.

kyukhin:
All relevant issues seems to be resolved. So, removing `blocked` label.

oleg-jukovec:
It seems that we already have tests for this cases:

https://github.com/tarantool/expirationd/blob/b52298fef4b6b092c4db6910ee3c41f11494e7b1/test/unit/space_index_test.lua

oleg-jukovec:
It seems that we already have tests for this cases:

https://github.com/tarantool/expirationd/blob/b52298fef4b6b092c4db6910ee3c41f11494e7b1/test/unit/space_index_test.lua

PR #67, opened 2021-07-26
WIP: Role

ArtDu:
Role to support saving the expirationd tasks state in the form of start_key, a key that was not expired before the task was turned off. This key can be obtained to the user using cartridge.vars, as shown in the test.


vrogach2020:
- create ticket with RFC
- describe role and get approves from repo owners

ligurio:
> After discussion with @ligurio and @ArtDu there are some follow-ups:

So I'll close PR.

PR #70, opened 2021-09-08
Fix freezes when stopping a task

### LINKED ISSUE #69: Task stop(restart) may freeze up to task.full_scan_delay seconds
We've encountered expd freezing on 
https://github.com/tarantool/expirationd/blob/930c1c4852cb6eea6ac32bb758753ad6d9bd71f3/expirationd.lua#L170

This happens due to worker_fiber being sleepy (so cancel fails), and fiber calling task.start got stuck.
This happened on production, led to malfunctioning cluster, pretty nasty.
Quickfix would be to add fiber.testcancel() after pcall:
https://github.com/tarantool/expirationd/blob/930c1c4852cb6eea6ac32bb758753ad6d9bd71f3/expirationd.lua#L152

I can provide more details regarding reproduceer/test-case, if necessary.

ArtDu:
The problem is that the call to task:stop hangs, because task:stop
sends cancel to the active fibers and waits for these fibers to become
"dead". But there may be some code that can get cancelation using pcall
and skip it, so we have to check the fiber for cancellation.
We do this by checking a flag `worker_canceled` that was previously
only used for atomic iterations. We also fixed the places where this
flag is checked and added comments why in this or that place we
check the cancellation of the worker.

Closes #69

### DIFF
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -102,6 +102,11 @@ local function default_do_worker_iteration(task)
     local checked_tuples_count = 0
     local vinyl_checked_tuples_count = 0
     if task.atomic_iteration then
+        -- Check before starting the transaction,
+        -- since a transaction can be long.
+        if task.worker_cancelled then
+            return true
+        end
         box.begin()
     end
     for _, tuple in task:iterate_with() do
@@ -113,6 +118,7 @@ local function default_do_worker_iteration(task)
         if checked_tuples_count >= task.tuples_per_iteration then
             if task.atomic_iteration then
                 box.commit()
+                -- The suspend functions can be long.
                 if task.worker_cancelled then
                     return true
                 end
@@ -127,6 +133,11 @@ local function default_do_worker_iteration(task)
                 suspend(task)
             end
             if task.atomic_iteration then
+                -- Check before starting the transaction,
+                -- since a transaction can be long.
+                if task.worker_cancelled then
+                    return true
+                end
                 box.begin()
             end
         end
@@ -144,12 +155,12 @@ local function worker_loop(task)
     fiber.name(string.format("worker of %q", task.name), { truncate = true })
 
     while true do
-        if task.worker_cancelled then
-            fiber.self():cancel()
-        end
         if (box.cfg.replication_source == nil and box.cfg.replication == nil) or task.force then
             task.on_full_scan_start(task)
             local state, err = pcall(task.do_worker_iteration, task)
+            -- Following functions are on_full_scan*,
+            -- but we probably did not complete the full scan,
+            -- so we should check for cancellation here.
             if task.worker_cancelled then
                 fiber.self():cancel()
             end
@@ -166,6 +177,12 @@ local function worker_loop(task)
             end
         end
 
+        -- If we do not check the fiber for cancellation,
+        -- then the fiber may fall asleep for a long time, depending on `full_scan_delay`.
+        -- And a fiber that wants to stop this task can also freeze, a kind of deadlock.
+        if task.worker_cancelled then
+            fiber.self():cancel()
+        end
         -- Full scan iteration is complete, yield
         fiber.sleep(task.full_scan_delay)
     end
@@ -212,9 +229,8 @@ local Task_methods = {
             self.guardian_fiber = nil
         end
         if (get_fiber_id(self.worker_fiber) ~= 0) then
-            if self.atomic_iteration then
-                self.worker_cancelled = true
-            else
+            self.worker_cancelled = true
+            if not self.atomic_iteration then
                 self.worker_fiber:cancel()
             end
             while self.worker_fiber:status() ~= "dead" do
--- a/test/unit/task_stop_test.lua
+++ b/test/unit/task_stop_test.lua
@@ -1,0 +1,34 @@
+local expirationd = require("expirationd")
+local fiber = require("fiber")
+local t = require("luatest")
+local g = t.group("task_stop")
+
+local helpers = require("test.helper")
+
+g.before_all(function()
+    helpers.init_spaces(g)
+end)
+
+g.after_each(function()
+    helpers.truncate_spaces(g)
+end)
+
+function g.test_cancel_on_pcall()
+    local function on_full_scan_complete()
+        pcall(fiber.sleep, 1)
+    end
+    local one_hour = 3600
+    local task = expirationd.start("clean_all", g.tree.id, helpers.is_expired_true, {
+        full_scan_delay = one_hour,
+        on_full_scan_complete = on_full_scan_complete
+    })
+    helpers.retrying({}, function()
+        t.assert(task.worker_fiber)
+    end)
+    -- We need to execute in a separate fiber,
+    -- since pcall does not check testcancel and stop may freeze up.
+    local f = fiber.create(task.stop, task)
+    helpers.retrying({timeout = 5}, function()
+        t.assert_equals(f:status(), "dead")
+    end)
+end


expirationd.lua
```diff
@@ -166,6 +166,8 @@ local function worker_loop(task)
             end
         end
 
+        --check if fiber was canceled in pcall
+        fiber.testcancel()
```
olegrok, line 170: Please add a space between `--` and message. Also please start a comment with a capital letter.

test/unit/stop_test.lua
```diff
@@ -0,0 +1,34 @@
...
+    helpers.init_spaces(g)
+end)
+
+g.after_each(function()
+    helpers.truncate_spaces(g)
+end)
+
+function g.test_cancel_on_pcall()
+    local function on_full_scan_complete()
+        pcall(function() fiber.sleep(1) end)
```
olegrok, line 18: nit: Could be rewritten as pcall(fiber.sleep, 1)

olegrok (commented):
Thanks for your patch, please add more info to commit message and remove `:` after `Closes`.

Also consider my comments below.

test/unit/stop_test.lua
```diff
@@ -0,0 +1,34 @@
...
+
+    local task = expirationd.start("clean_all", g.tree.id, helpers.is_expired_true, {
+        full_scan_delay = 123123123,
+        on_full_scan_complete = on_full_scan_complete
+    })
+    helpers.retrying({}, function()
+        t.assert(task.worker_fiber)
+    end)
+    -- we need to execute in a separate fiber,
+    -- since pcall does not check testcancel and stop may freeze up
```
ligurio, line 29: Sentence starts with uppercase and dot at the end.
https://www.tarantool.io/en/doc/latest/dev_guide/lua_style_guide/

ligurio (changes requested):
Thanks for the patch! Please add description or explanation of changes to a commit message.

expirationd.lua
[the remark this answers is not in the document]
ArtDu: Yes, you are right, I reviewed how these checks are arranged and redid it a bit. I also left comments on each check, about why it should be here. Please look and let's discuss what you think is the best thing to do?

ArtDu: Sorry, I missed the space again ((

test/unit/stop_test.lua
ArtDu: Yes, that would be more correct. Fixed

test/unit/stop_test.lua
```diff
@@ -0,0 +1,34 @@
...
+    helpers.truncate_spaces(g)
+end)
+
+function g.test_cancel_on_pcall()
+    local function on_full_scan_complete()
+        pcall(function() fiber.sleep(1) end)
+    end
+
+    local task = expirationd.start("clean_all", g.tree.id, helpers.is_expired_true, {
+        full_scan_delay = 123123123,
```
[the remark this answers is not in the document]
ArtDu: Was rewrite to `local one_hour = 3600`

expirationd.lua
```diff
@@ -166,6 +166,8 @@ local function worker_loop(task)
             end
         end
 
+        --check if fiber was canceled in pcall
```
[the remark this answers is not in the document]
ArtDu: Yeah, fixed

test/unit/stop_test.lua
ArtDu: Sorry, but I could not find in the guideline which you attached the words uppercase and dot, perhaps we need to leave a ticket to the documentation. And we will probably need to create a ticket in `expd` to redo the old comments so that they also match the comments style-guide

expirationd.lua
```diff
@@ -102,6 +102,11 @@ local function default_do_worker_iteration(task)
     local checked_tuples_count = 0
     local vinyl_checked_tuples_count = 0
     if task.atomic_iteration then
+        -- Before starting the transaction,
+        -- since the batch transaction can be long.
```
olegrok, line 106: Sorry, but I don't understand that "batch transaction" means.

expirationd.lua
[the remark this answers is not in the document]
olegrok: To be honest it looks a bit tricky. Ideally each line should be covered by test. But I guess it's not so.
Anyway "Check if fiber was canceled" doesn't explain why above we use "task.worker_canceled" but here we use testcancel?

expirationd.lua
```diff
@@ -144,12 +155,12 @@ local function worker_loop(task)
...
     while true do
-        if task.worker_canceled then
-            fiber.self():cancel()
-        end
         if (box.cfg.replication_source == nil and box.cfg.replication == nil) or task.force then
             task.on_full_scan_start(task)
             local state, err = pcall(task.do_worker_iteration, task)
+            -- Following functions are on_full_scan*,
+            -- but we probably did not complete the full scan,
+            -- so we should logically check cancellation here.
```
olegrok, line 163: It's a bit strange phrase.

Could we perform "non-logical" check? Also I guess it should be "check for cancellation".

olegrok (commented):
Thanks for your fixes. See my comments below.

expirationd.lua
ArtDu: Can I replace it with transaction per batch?

olegrok: why not just "transaction"?

ArtDu: Yeah, we can write just a transaction)

test/unit/stop_test.lua
ligurio: https://github.com/tarantool/doc/issues/2331

expirationd.lua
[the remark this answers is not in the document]
ArtDu: We use both test_cancel and task.worker_canceled here.
Check this out, is this a good description?
```diff
@@ -177,11 +177,15 @@ local function worker_loop(task)
             end
         end

-        -- Check if fiber was canceled.
-        fiber.testcancel()
-        -- Then there is sleep and it can be long.
+        -- Check cancellation of fiber,
+        -- then there is sleep and it can be long.
         if task.worker_canceled then
+            -- The worker_canceled flag is used only if we have atomic_iteration = true.
             fiber.self():cancel()
+        else
+            -- Check if fiber was canceled, because pcall
+            -- can catch a cancellation and can skip it.
+            fiber.testcancel()
         end
         -- Full scan iteration is complete, yield
         fiber.sleep(task.full_scan_delay)
```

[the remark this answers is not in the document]
olegrok: > -- Check if fiber was canceled, because pcall
            -- can catch a cancellation and can skip it.
            fiber.testcancel()

Who cancelled this fiber? And why it doesn't switch worker_canceled flag to true?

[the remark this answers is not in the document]
ArtDu: > Who canceled this fiber? And why it doesn't switch worker_canceled flag to true?  
 
We can write about the task:stop logic, but why describe what you can see in the code?
I missed your point, what do you want to see here as comments?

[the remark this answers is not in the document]
olegrok: Seems I told about this part:
https://github.com/tarantool/expirationd/blob/930c1c4852cb6eea6ac32bb758753ad6d9bd71f3/expirationd.lua#L215-L219

It's a bit inconsistent. I thought that stop() turn worker_canceled flag on anyway. But seems it's relevant only for atomic_iteration.

I don't know will anyone support me, but I vote for setup `worker_canceled = true` for not batch_iteration case as well. Seems it also should solve an issue.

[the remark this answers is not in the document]
ArtDu: Yes, this is a good option, then we don't need testcancel check at all. Perhaps it is worth doing so. @ligurio can help discuss it as a third person

expirationd.lua
ArtDu: Did this:
```diff
             local state, err = pcall(task.do_worker_iteration, task)
             -- Following functions are on_full_scan*,
             -- but we probably did not complete the full scan,
-            -- so we should logically check cancellation here.
+            -- so we should check for cancellation here.
             if task.worker_canceled then
                 fiber.self():cancel()
             end
```

expirationd.lua
ArtDu: Did this:
```diff
@@ -103,7 +103,7 @@ local function default_do_worker_iteration(task)
     local vinyl_checked_tuples_count = 0
     if task.atomic_iteration then
         -- Before starting the transaction,
-        -- since the batch transaction can be long.
+        -- since the transaction can be long.
         if task.worker_canceled then
             return true
         end
```

expirationd.lua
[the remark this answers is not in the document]
ligurio: Personally I like a variant with `worker_canceled = true` for non-`batch_iteration` as well.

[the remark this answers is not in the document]
ArtDu: Finished with this. Changed worker_canceled not only for atomic and removed testcancel.

test/unit/stop_test.lua
```diff
@@ -0,0 +1,34 @@
...
+    })
+    helpers.retrying({}, function()
+        t.assert(task.worker_fiber)
+    end)
+    -- We need to execute in a separate fiber,
+    -- since pcall does not check testcancel and stop may freeze up.
+    local f = fiber.create(task.stop, task)
+    helpers.retrying({}, function()
+        t.assert_equals(f:status(), "dead")
+    end)
```
ligurio, line 33: The fixed bug is about freeze on stop, right now in test you just check that fiber some day finished it's work. But we need to check that fiber will finish during  some time. I would suggest to set explicit timeout to `helpers.retrying()`.

expirationd.lua
```diff
@@ -127,6 +133,11 @@ local function default_do_worker_iteration(task)
                 suspend(task)
             end
             if task.atomic_iteration then
+                -- Before starting the transaction,
+                -- since the batch transaction can be long.
```
olegrok, line 137: Seems "batch" is redundant here

expirationd.lua
```diff
@@ -102,6 +102,11 @@ local function default_do_worker_iteration(task)
     local checked_tuples_count = 0
     local vinyl_checked_tuples_count = 0
     if task.atomic_iteration then
+        -- Before starting the transaction,
```
olegrok, line 105: "Check before starting transaction, ..." or something similar a phase "Before starting the transaction, since the transaction can be long" without any verb doesn't make sense.

ArtDu: You are right, fixed on "Check before starting transaction, ..."

expirationd.lua
ArtDu: Yes, deleted that word

expirationd.lua
```diff
@@ -166,6 +177,10 @@ local function worker_loop(task)
             end
         end
 
+        -- There is sleep and it can be long.
```
olegrok, line 180: Yes, it can be long. Please add more context to this comment. Currently it's a bit useless.

ligurio (changes requested):

test/unit/stop_test.lua
ArtDu: There is a default timeout in the helpers.retrying(), 5 seconds. Or do you want it to be visible right in the code?

expirationd.lua
```diff
@@ -127,6 +133,11 @@ local function default_do_worker_iteration(task)
                 suspend(task)
             end
             if task.atomic_iteration then
+                -- Before starting the transaction,
```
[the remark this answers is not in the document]
ArtDu: Sorry, my fault

expirationd.lua
ArtDu: That's ok? 
```lua
-- If we do not check the fiber for cancellation,
-- then the fiber may fall asleep for a long time, depending on `full_scan_delay`.
-- And a fiber that wants to stop this task can also freeze, a kind of deadlock.
```

test/unit/stop_test.lua
ligurio: Yep, let's make it at least explicit.

olegrok (approved):

test/unit/stop_test.lua
ArtDu: Made explicit, 5 seconds

test/unit/stop_test.lua
```diff
@@ -0,0 +1,34 @@
+local expirationd = require("expirationd")
```
ligurio, line 1: Could you please rename file with test to something like `task_test.lua` or `task_stop_test.lua`. Right now it is unclear what kind of stop we test here without looking to a file content. We plan to move TAP tests to luatest format and it would be good to give a clear names for files.

ligurio (changes requested):

test/unit/stop_test.lua
ArtDu: Renamed to `task_stop_test.lua`

ligurio (approved):

ligurio:
LGTM

PR #71, opened 2021-09-09
Fixes in Makefile and expirationd.lua

ligurio:


Makefile
```diff
@@ -9,10 +10,10 @@ luacheck:
 .PHONY: test
 test:
 	.rocks/bin/luatest -v
-	rm -rf *.xlog* *.snap 51{2,3,4,5,6,7}
+	rm -rf ${CLEANUP_FILES}
 	INDEX_TYPE='TREE' SPACE_TYPE='vinyl' ./test.lua
-	rm -rf *.xlog* *.snap 51{2,3,4,5,6,7}
+	rm -rf ${CLEANUP_FILES}
 	INDEX_TYPE='HASH' ./test.lua
-	rm -rf *.xlog* *.snap 51{2,3,4,5,6,7}
```
olegrok, line 16: `51{2,3,4,5,6,7}` is about directories that vinyl creates. Why aren't they in CLEANUP_FILES?

ligurio: I thought it is a set of files for sophia or other legacy.

olegrok (approved):

Makefile
```diff
@@ -1,3 +1,8 @@
+CLEANUP_FILES  = tarantool.log
+CLEANUP_FILES += *.xlog*
```
Totktonada, line 2: 1. A typo (`*.xlog*` -> `*.xlog`)?
2. How about `*.vylog`?

README.md
```diff
@@ -9,6 +9,23 @@ tuple and performs the expiry itself: either deletes it (memcache), or
 does something smarter, like put a smaller representation of the data
 being deleted into some other space.
 
+There are a number of similar modules:
+- [moonwalker](https://github.com/tarantool/moonwalker) triggered manually,
+useful for batch transactions, a performance about 600/700k rec/sec
+- [expirationd](https://github.com/tarantool/expirationd/issues/53) always
+expires tuples without using indices but using any condition, without guarantee
+for time expiration.
```
Totktonada, line 17: > always expires tuples without using indices

Now it is not quite correct, right? See #50 and PR #52.

Makefile
ligurio: I kept all file masks as is, but moved them to a separate variable. However 1 and 2 are good points. I think we can address it in #76. Vinyl tests are broken and I think it's a reason why `*.vylog` is not there.

README.md
ligurio: Probably yes. Although proposed table is good enough to make an opinion about these projects.

Issue #14, opened 2016-08-01
Document usage on replica and  `force` option more clearly


rtsisyk:
I and @kostja spent some time on IoT Hackathon trying to get expirationd working. It would be nice to log a warning that expirationd doesn't work on replica by default and `force` option should be used.

bigbes:
Document vinyl-specific options and tell about specific of vinyl expiration.

### Options:

```
+    -- assumed size of vinyl space (in the first iteration)
+    vinyl_assumed_space_len = math.pow(10, 7),
+    -- factor for recalculation of vinyl space size
+    vinyl_assumed_space_len_factor = 2
```

### Specific:

Vinyl space size can't be counted (since many operations, `upsert` for example, are applied when you address some data), so you should count (approximate space size) tuples with the first start. `vinyl_assumed_space_len` is approximate count for first run and `vinyl_assumed_space_len_factor` for next milestone (after we've reached <assumed space size> next milestone is <factor>*<assumed space size> and so on)

PR #72, opened 2021-09-09
Support LDoc

ligurio:
Online: https://tarantool.github.io/expirationd/

Ticket: #79

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,22 @@
+# This way everything works as expected ever for
+# `make -C /path/to/project` or
+# `make -f /path/to/project/Makefile`.
+MAKEFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
+PROJECT_DIR := $(patsubst %/,%,$(dir $(MAKEFILE_PATH)))
+
 CLEANUP_FILES  = tarantool.log
 CLEANUP_FILES += *.xlog*
 CLEANUP_FILES += *.snap
 CLEANUP_FILES += 51{2,3,4,5,6,7}  #  Directories that vinyl creates.
 
 all: test
+
+# The template (ldoc.tpl) is written using tarantool specific
+# functions like string.split(), string.endswith(), so we run
+# ldoc using tarantool.
+apidoc:
+	.rocks/bin/ldoc -c $(PROJECT_DIR)/doc/ldoc/config.ld \
+             -d $(PROJECT_DIR)/doc/apidoc/ expirationd.lua
 
 check: luacheck
 
--- a/doc/ldoc/assets/ldoc.ltp
+++ b/doc/ldoc/assets/ldoc.ltp
@@ -1,0 +1,420 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
+   "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
+<html>
+<meta http-equiv="Content-Type" content="text/html; charset=$(ldoc.doc_charset)"/>
+<head>
+    <title>$(ldoc.title)</title>
+    <link rel="stylesheet" href="$(ldoc.css)" type="text/css" />
+# if ldoc.custom_css then -- add custom CSS file if configured.
+    <link rel="stylesheet" href="$(ldoc.custom_css)" type="text/css" />
+# end
+</head>
+<body>
+
+<div id="container">
+
+<div id="product">
+	<div id="product_logo"></div>
+	<div id="product_name"><big><b></b></big></div>
+	<div id="product_description"></div>
+</div> <!-- id="product" -->
+
+
+<div id="main">
+
+# local no_spaces = ldoc.no_spaces
+# local use_li = ldoc.use_li
+# local display_name = ldoc.display_name
+# local iter = ldoc.modules.iter
+# local function M(txt,item) return ldoc.markup(txt,item,ldoc.plain) end
+# local nowrap = ldoc.wrap and '' or 'nowrap'
+
+# local function parse_source_file_table(filename, start_lineno)
+#     local content = ldoc.include_file(filename)
+#     local skip = true
+#     local res = {}
+#     for lineno, line in ldoc.ipairs(content:split('\n')) do
+#         if lineno == start_lineno then
+#             skip = false
+#             goto continue
+#         end
+#         if not skip and line:endswith('}') then
+#             return res
+#         end
+#         if not skip then
+#             local key, value = line:match("^ *([%w_]-) *= *([%w']-),?$")
+#             if key then
+#                 res[key] = value
+#             end
+#         end
+#         ::continue::
+#     end
+#     return res
+# end
+
+# -- Copied from config.ld.
+# --
+# -- A section description is not processed by
+# -- `custom_display_name_handler`.
+# local function convert_markdown_references(text)
+#     local refs = {}
+#     for _, line in ldoc.pairs(text:split('\n')) do
+#        local anchor, url = line:lstrip():match('^%[([0-9])%]: (.*)$')
+#        if anchor then
+#           refs[anchor] = url
+#        end
+#     end
+#
+#     for anchor, url in ldoc.pairs(refs) do
+#         text = text:gsub('\n *%[' .. anchor .. '%]: [^\n]*', '')
+#         text = text:gsub('%[(.-)%]%[' .. anchor .. '%]',
+#             '<a href="' .. refs[anchor] .. '">%1</a>')
+#     end
+#     return text
+# end
+
+# -- Copied from config.ld.
+# --
+# -- A section description is not processed by
+# -- `custom_display_name_handler`.
+# local function convert_markdown_backticks(text)
+#     return text:gsub('`(.-)`', '<code>%1</code>')
+# end
+
+<!-- Menu -->
+
+<div id="navigation">
+<br/>
+<h1>$(ldoc.project)</h1>
+
+# if not ldoc.single and module then -- reference back to project index
+<ul>
+  <li><a href="../$(ldoc.output).html">Index</a></li>
+</ul>
+# end
+
+# --------- contents of module -------------
+# if module and not ldoc.no_summary and #module.items > 0 then
+<h2>Contents</h2>
+<ul>
+# for kind,items in module.kinds() do
+<li><a href="#$(no_spaces(kind))">$(kind)</a></li>
+# end
+</ul>
+# end
+
+
+# if ldoc.no_summary and module and not ldoc.one then -- bang out the functions on the side
+# for kind, items in module.kinds() do
+<h2>$(kind)</h2>
+<ul class="nowrap">
+# for item in items() do
+    <li><a href="#$(item.name)">$(display_name(item))</a></li>
+# end
+</ul>
+# end
+# end
+# -------- contents of project ----------
+# local this_mod = module and module.name
+# for kind, mods, type in ldoc.kinds() do
+#  if ldoc.allowed_in_contents(type,module) then
+<h2>$(kind)</h2>
+<ul class="$(kind=='Topics' and '' or 'nowrap')">
+#  for mod in mods() do local name = display_name(mod)
+#   if mod.name == this_mod then
+  <li><strong>$(name)</strong></li>
+#   else
+  <li><a href="$(ldoc.ref_to_module(mod))">$(name)</a></li>
+#   end
+#  end
+# end
+</ul>
+# end
+
+</div>
+
+<div id="content">
+
+# if ldoc.body then -- verbatim HTML as contents; 'non-code' entries
+    $(ldoc.body)
+# elseif module then -- module documentation
+<h1>$(ldoc.module_typename(module)) <code>$(module.name)</code></h1>
+<p>$(M(module.summary,module))</p>
+<p>$(M(module.description,module))</p>
+#   if module.tags.include then
+        $(M(ldoc.include_file(module.tags.include)))
+#   end
+#   if module.see then
+#     local li,il = use_li(module.see)
+    <h3>See also:</h3>
+    <ul>
+#     for see in iter(module.see) do
+         $(li)<a href="$(ldoc.href(see))">$(see.label)</a>$(il)
+#    end -- for
+    </ul>
+#   end -- if see
+#   if module.usage then
+#     local li,il = use_li(module.usage)
+    <h3>Usage:</h3>
+    <ul>
+#     for usage in iter(module.usage) do
+        $(li)<pre class="example">$(ldoc.escape(usage))</pre>$(il)
+#     end -- for
+    </ul>
+#   end -- if usage
+#   if module.info then
+    <h3>Info:</h3>
+    <ul>
+#     for tag, value in module.info:iter() do
+        <li><strong>$(tag)</strong>: $(M(value,module))</li>
+#     end
+    </ul>
+#   end -- if module.info
+
+
+# if not ldoc.no_summary then
+# -- bang out the tables of item types for this module (e.g Functions, Tables, etc)
+# for kind,items in module.kinds() do
+<h2><a href="#$(no_spaces(kind))">$(kind)</a></h2>
+<table class="function_list">
+#  for item in items() do
+	<tr>
+	<td class="name" $(nowrap)><a href="#$(item.name)">$(display_name(item))</a></td>
+	<td class="summary">$(M(item.summary:split('\n', 1)[1],item))</td>
+	</tr>
+#  end -- for items
+</table>
+#end -- for kinds
+
+<br/>
+<br/>
+
+#end -- if not no_summary
+
+# --- currently works for both Functions and Tables. The params field either contains
+# --- function parameters or table fields.
+# local show_return = not ldoc.no_return_or_parms
+# local show_parms = show_return
+# for kind, items in module.kinds() do
+#   local kitem = module.kinds:get_item(kind)
+#   local has_description = kitem and ldoc.descript(kitem) ~= ""
+    <h2 class="section-header $(has_description and 'has-description')"><a name="$(no_spaces(kind))"></a>$(kind)</h2>
+    $(M(module.kinds:get_section_description(kind),nil))
+#   if kitem then
+#       if has_description then
+          <div class="section-description">
+#         local item_text = ldoc.descript(kitem)
+#         item_text = convert_markdown_references(item_text)
+#         item_text = convert_markdown_backticks(item_text)
+          $(M(item_text,kitem))
+          </div>
+#       end
+#       if kitem.usage then
+            <h3>Usage:</h3>
+            <pre class="example">$(ldoc.prettify(kitem.usage[1]))</pre>
+#        end
+#   end
+    <dl class="function $(kitem and kitem.name or '')">
+#  for item in items() do
+    <dt>
+    <a name="$(item.name)"></a>
+#   if kitem and (kitem.name == 'options' or kitem.name == 'defaults') then
+    <strong>$(item.summary)</strong>
+#   else
+    <strong>$(display_name(item))</strong>
+#   end
+#   if ldoc.prettify_files and ldoc.is_file_prettified[item.module.file.filename] then
+    <a style="float:right;" href="$(ldoc.source_ref(item))">line $(item.lineno)</a>
+#  end
+    </dt>
+    <dd>
+#   if kitem and (kitem.name == 'options' or kitem.name == 'defaults') then
+    $(M(item.description))</p>
+#   else
+    <p>$(M(ldoc.descript(item),item))</p>
+#   end
+
+#   if ldoc.custom_tags then
+#    for custom in iter(ldoc.custom_tags) do
+#     local tag = item.tags[custom[1]]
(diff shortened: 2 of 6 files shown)


expirationd.lua
```diff
@@ -310,48 +306,129 @@ end
...
+--     to process tuples from each batch in a single transaction.
+-- @boolean[opt] options.force
+--     expirationd doesn't work on replica by default. With option `force` it
+--     should be possible to force work on replica.
+-- @number[opt] options.full_scan_delay
+--     Sleep time between full scans (in seconds). It is allowed to pass an FFI
+--     number: `1LL`, `1UL` etc. Default value is 1 sec.
+-- @number[opt] options.full_scan_time
+--     Time required for a full index scan (in seconds). It is allowed to pass
+--     an FFI number: 1LL, 1UL etc. Default value is 3600.
```
ArtDu, line 341: The time required for a full scan of the index, but what happens if the real full scan time is longer than the specified time, or vice versa, faster? How does this field affect the logic of the entire work of the task? The logic for interacting with this field and others is partially described in this ticket #38, but not very obvious.

expirationd.lua
```diff
@@ -192,12 +183,17 @@ end
 -- Task management
 -- ------------------------------------------------------------------------- --
 
+--[[
+task:start() - Force start `task` (old `guardian_fiber` will be garbage
+               collected, so do `task:stop()` before or `task:restart()` instead)
```
ArtDu, line 188: I know that the documentation for task: api is not ready yet. But I will comment on what you have already written.
> old `guardian_fiber` will be garbage collected, so do` task:stop()` before or` task:restart()` instead

I did not understand the point, it also calls `task:stop` inside and waits for the previous task to stop.

expirationd.lua
```diff
@@ -310,48 +306,129 @@ end
...
+-- @string name
+--     Task name.
+-- @string space_id
+--     Space to look in for expired tuples.
+-- @func is_tuple_expired
+--     Function, must accept tuple and return true/false (is tuple expired or
+--     not), receives `args`, `tuple` as arguments.
+-- @table[opt] options
+--     Table with named options.
+-- @table[opt] options.args
```
ArtDu, line 327: Perhaps it is worth putting, at the beginning, the parameters that have a stronger effect on the customization of the task. For example, the index is often searched for and asked in the public Tarantool chat

ligurio: Options sorted alphabetically order by name. It's allow to find parameters that have strong effect on the customization personally for you ;)

expirationd.lua
ligurio: > I know that the documentation for task: api is not ready yet. 

This PR is not about changing API, but about documenting it right and precisely.

> But I will comment on what you have already written.

It's text that was written before in comments, I just formatted it as LDoc comment. However I'll check it this still valid.

ligurio: Removed.

expirationd.lua
```diff
@@ -327,48 +390,201 @@ end
...
+-- - Process min(`space_length`, `tuples_per_iteration`) tuples at once.
+--
+-- - Sleep `tuples_per_iteration` × `full_scan_time` / `space_length` (but not beyond 1 second).
+--
+-- - Repeat 1-2 until the whole space will be traversed.
+--
+-- - Sleep 1 second.
+--
+-- - Repeat 1-4.
+--
```
ArtDu, line 414: Maybe use numbers instead of dots. It's clearer with repeat 1-2/1-4?
Now:
image

Numbers:
image

doc/ldoc/assets/ldoc.ltp
```diff
@@ -0,0 +1,420 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
```
olegrok, line 1: Do we really need to push this generated code? I guess CI should push it github.io or at least we should have protection when we change something and don't update a doc.

ligurio: it's a template used by ldoc:

```
sergeyb@pony:~/sources/MRG/expirationd$ make apidoc 
.rocks/bin/ldoc -c /home/sergeyb/sources/MRG/expirationd/doc/ldoc/config.ld \
             -d /home/sergeyb/sources/MRG/expirationd/doc/apidoc/ expirationd.lua
reading configuration from /home/sergeyb/sources/MRG/expirationd/doc/ldoc/config.ld
template not found at 'doc/ldoc/assets' Use -l to specify directory containing ldoc.ltp
make: *** [Makefile:19: apidoc] Error 255
```

expirationd.lua
ligurio: Sure, fixed.

```lua
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -402,15 +402,16 @@ end
 --
 -- How expirationd works in general:
 --
--- - Process min(`space_length`, `tuples_per_iteration`) tuples at once.
+-- 1. Process min(`space_length`, `tuples_per_iteration`) tuples at once.
 --
--- - Sleep `tuples_per_iteration` × `full_scan_time` / `space_length` (but not beyond 1 second).
+-- 2. Sleep `tuples_per_iteration` × `full_scan_time` / `space_length` (but not
+--    beyond 1 second).
 --
--- - Repeat 1-2 until the whole space will be traversed.
+-- 3. Repeat 1-2 until the whole space will be traversed.
 --
--- - Sleep 1 second.
+-- 4. Sleep 1 second.
 --
--- - Repeat 1-4.
+-- 5. Repeat 1-4.
 --
 --
 -- @string name
```

expirationd.lua
```diff
@@ -327,48 +390,202 @@ end
...
+-- ```
+-- local function is_tuple_expired(args, tuple)
+--     local tuple_expire_time = get_field(tuple, args.field_no)
+--     if type(tuple_expire_time) ~= 'number' then
+--         return true
+--     end
+--     local current_time = fiber.time()
+--
+--     return current_time >= tuple_expire_time
+-- end
```
olegrok, line 436: image

looks broken

olegrok: image

functions below looks fine.

Makefile
```diff
@@ -1,10 +1,24 @@
+# This way everything works as expected ever for
+# `make -C /path/to/project` or
+# `make -f /path/to/project/Makefile`.
+MAKEFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
+PROJECT_DIR := $(patsubst %/,%,$(dir $(MAKEFILE_PATH)))
+LUACOV_REPORT := $(PROJECT_DIR)/luacov.report.out
```
Totktonada, line 6: Luacov seems irrelevant here.

expirationd.lua
```diff
@@ -327,48 +390,202 @@ end
 -- ========================================================================= --
 -- Expiration daemon management functions
 -- ========================================================================= --
+--
+-- {{{ Module methods
+--
+
+--- Module methods
+--
+-- @section Functions
```
Totktonada, line 399: In my understanding a method accepts an argument like `self` or `this`. So I think that 'functions' would be more appropriate term for those pieces of code.

expirationd.lua
```diff
@@ -327,48 +390,202 @@ end
...
+-- 1. Process min(`space_length`, `tuples_per_iteration`) tuples at once.
+--
+-- 2. Sleep `tuples_per_iteration` × `full_scan_time` / `space_length` (but not
+--    beyond 1 second).
+--
+-- 3. Repeat 1-2 until the whole space will be traversed.
+--
+-- 4. Sleep 1 second.
+--
+-- 5. Repeat 1-4.
```
Totktonada, line 414: 1. Looks relevant to #38.
2. It is rendered as unnumbered list on https://tarantool.github.io/expirationd/, but when I generated it locally, the list is numbered. I guess the GitHub Pages webpage contains not a last state.

expirationd.lua
```diff
@@ -327,48 +390,202 @@ end
...
+--
+--
+-- @string name
+--     Task name.
+-- @string space_id
+--     Space to look in for expired tuples.
```
Totktonada, line 420: A string or numerical ID, correct?

expirationd.lua
```diff
@@ -327,48 +390,202 @@ end
...
+--     expired or not), receives `args` and `tuple` as arguments.
+--
+-- Example of function:
+--
+-- ```
+-- local function is_tuple_expired(args, tuple)
+--     local tuple_expire_time = get_field(tuple, args.field_no)
+--     if type(tuple_expire_time) ~= 'number' then
+--         return true
+--     end
```
Totktonada, line 432: I don't spot the idea of this check. If there is no expire information within a tuple, expire it anyway?

expirationd.lua
```diff
@@ -327,48 +390,202 @@ end
...
+--     Table with named options.
+-- @table[opt] options.args
+--     Passed to `is_tuple_expired()` and `process_expired_tuple()` as
+--     an additional context.
+-- @boolean[opt] options.atomic_iteration
+--     False (default) to process each tuple as a single transaction and true
+--     to process tuples from each batch in a single transaction.
+-- @boolean[opt] options.force
+--     expirationd doesn't work on replica by default. With option `force` it
+--     should be possible to force work on replica.
```
Totktonada, line 449: It deserves a top level note (after this we can say that #14 is resolved). I heard 'why expirationd does not work' so many times...

I would also clarify what exactly 'replica' means in this context.

expirationd.lua
```diff
@@ -327,48 +390,202 @@ end
...
+-- @number[opt] options.full_scan_delay
+--     Sleep time between full scans (in seconds). It is allowed to pass an FFI
+--     number: `1LL`, `1UL` etc. Default value is 1 sec.
```
Totktonada, line 452: Typo: 1UL -> 1ULL.

expirationd.lua
```diff
@@ -327,48 +390,202 @@ end
...
+-- @number[opt] options.full_scan_time
+--     Time required for a full index scan (in seconds). It is allowed to pass
+--     an FFI number: `1LL`, `1UL` etc. `full_scan_time` used for calculation
```
Totktonada, line 455: Typo: 1UL -> 1ULL.

expirationd.lua
```diff
@@ -327,48 +390,202 @@ end
...
+-- Example of function:
+--
+-- ```
+-- local function iterate_with()
+--     return true
+-- end
+-- ```
+-- @number[opt] options.iteration_delay
+--     Max sleep time between iterations (in seconds). It is allowed to pass
+--     an FFI number: `1LL`, `1UL` etc. Default value is 1 sec.
```
Totktonada, line 477: Typo: 1UL -> 1ULL.

expirationd.lua
```diff
@@ -327,48 +390,202 @@ end
...
+--     Function which returns an iterator object which provides tuples to
+--     check, considering the `start_key`, `process_while` and other options.
+--     There's a default function which can be overriden with this parameter.
+--
+-- Example of function:
+--
+-- ```
+-- local function iterate_with()
+--     return true
+-- end
```
Totktonada, line 473: Just above:

> Function, which returns an iterator object \<...\>

But this function does not return an iterator object (`gen`, `param` `state` triplet, where `gen` is callable).

Totktonada: And I don't see any information about arguments, which is passed to the function.

expirationd.lua
```diff
@@ -327,48 +390,202 @@ end
...
+--     Applied to expired tuples, receives `space_id`, `args`, `tuple` as
+--     arguments. When `process_expired_tuple` is not passed (or `nil` passed),
+--     tuples are removed.
+--
+-- Example of function:
+--
+-- ```
+-- local function put_tuple_to_archive(space_id, args, tuple)
+--     box.space[space_id]:delete{tuple[1]}
+--     local email = get_field(tuple, 2)
```
Totktonada, line 516: `tuple[1]` is used just above, but then `get_field(tuple, 2)`. Why? What is the difference?

expirationd.lua
```diff
@@ -327,48 +390,202 @@ end
...
+--
+-- Example of function:
+--
+-- ```
+-- local function process_while()
+--     return false
+-- end
+-- ```
+--
+-- @string[opt] options.start_key
```
Totktonada, line 536: A key may be of any time. I would also say that even if expirationd allows to pass a constant key there, there is no much sense to do so (just bad api design). Any real usage will pass a function there.

expirationd.lua
```diff
@@ -327,48 +390,202 @@ end
...
+--     Start iterating from the tuple with this index value. Or when iterator
+--     is 'EQ', iterate over tuples with this index value. Must be a value of
+--     the same data type as the index field or fields, or a function which
+--     returns such value. If omitted or nil, all tuples will be checked.
+-- @number[opt] options.tuples_per_iteration
+--     Number of tuples to check in one batch (iteration). It is allowed to
+--     pass an FFI number: `1LL`, `1UL` etc. Default value is 1024.
```
Totktonada, line 543: Typo: 1UL -> 1ULL.

expirationd.lua
```diff
@@ -327,48 +390,202 @@ end
...
+-- @number[opt] options.vinyl_assumed_space_len_factor
+--     Factor for recalculation of vinyl space size. Vinyl space size can't be
+--     counted (since many operations, `upsert` for example, are applied when
+--     you address some data), so you should count (approximate space size)
+--     tuples with the first start. `vinyl_assumed_space_len` is approximate
+--     count for first run and `vinyl_assumed_space_len_factor` for next
+--     milestone (after we've reached next milestone is `*` and so on). It is
+--     allowed to pass an FFI number: `1LL`, `1UL` etc. Default value is 2.
```
Totktonada, line 551: Same, 1UL -> 1ULL.

expirationd.lua
```diff
@@ -327,48 +390,202 @@ end
...
+-- @number[opt] options.vinyl_assumed_space_len
+--     Assumed size of vinyl space (in the first iteration).
+--     Vinyl space size can't be counted (since many operations, `upsert` for
+--     example, are applied when you address some data), so you should count
+--     (approximate space size) tuples with the first start.
+--     `vinyl_assumed_space_len` is approximate count for first run and
+--     `vinyl_assumed_space_len_factor` for next milestone (after we've reached
+--     next milestone is `*` and so on). It is allowed to pass an FFI number:
+--     `1LL`, `1UL` etc. Default value is 10^7.
```
Totktonada, line 560: Same, 1UL -> 1ULL.

expirationd.lua
```diff
@@ -327,48 +390,202 @@ end
...
+--
+-- @treturn task_obj
```
Totktonada, line 562: The link goes to the task function. Is it correct? I would expect that it'll go to the list of task object methods.

Makefile
ligurio: Right.

```
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,6 @@
 # `make -f /path/to/project/Makefile`.
 MAKEFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
 PROJECT_DIR := $(patsubst %/,%,$(dir $(MAKEFILE_PATH)))
-LUACOV_REPORT := $(PROJECT_DIR)/luacov.report.out
 
 CLEANUP_FILES  = tarantool.log
 CLEANUP_FILES += *.xlog*
```

expirationd.lua
ligurio: Fixed.

```lua
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -391,10 +391,10 @@ end
 -- Expiration daemon management functions
 -- ========================================================================= --
 --
--- {{{ Module methods
+-- {{{ Functions
 --
 
---- Module methods
+--- Functions
 --
 -- @section Functions
 
@@ -924,4 +924,4 @@ return {
     show_task_list = show_task_list_obsolete,
 }
 
--- }}} Module methods
+-- }}} Functions
```

expirationd.lua
ligurio: Right, fixed it.

expirationd.lua
ligurio: Fixed it. (copy-pasted from test)

```lua
 -- local function is_tuple_expired(args, tuple)
 --     local tuple_expire_time = get_field(tuple, args.field_no)
---     if type(tuple_expire_time) ~= 'number' then
---         return true
---     end
 --     local current_time = fiber.time()
---
 --     return current_time >= tuple_expire_time
 -- end
```

expirationd.lua
ligurio: Fixed.

expirationd.lua
ligurio: Fixed.

expirationd.lua
ligurio: Fixed.

expirationd.lua
ligurio: > A key may be of any time.

Type?


True. Below is a piece of code to process `start_key` argument:
```lua
    -- check start_key                                                           
    if options.start_key ~= nil or options.start_key == box.NULL then            
        if type(options.start_key) == "function" then                            
            task.start_key = function() return options.start_key() end           
        else                                                                     
            task.start_key = function() return options.start_key end             
        end                                                                      
    end      
```

Fixed a type of `start_key` in LDoc comment.

P.S. type of `start_key` in checks() is correct:
```lua
local function expirationd_run_task(name, space_id, is_tuple_expired, options)   
    checks('string', 'number|string', 'function', {                              
        args = '?',                                                              
        ...                                         
        start_key = '?',                                                         
        tuples_per_iteration = '?number|cdata',                                  
        vinyl_assumed_space_len_factor = '?number|cdata',                        
        vinyl_assumed_space_len = '?number|cdata',                               
    })                 
```

expirationd.lua
ligurio: Fixed.

expirationd.lua
ligurio: Fixed.

expirationd.lua
ligurio: Fixed,

expirationd.lua
ligurio: 2. Oleg already spotted it and I haven't update HTML version after fix.
Pushed latest version to https://tarantool.github.io/expirationd/.

expirationd.lua
ligurio: Removed example and instead described default function:
```lua
@@ -463,18 +460,25 @@ end
 -- @func[opt] options.iterate_with
 --     Function which returns an iterator object which provides tuples to
 --     check, considering the `start_key`, `process_while` and other options.
---     There's a default function which can be overriden with this parameter.
---
--- Example of function:
+--     When option is nil default function is used. Default function returns
+--     iterator returned by [index_object:pairs()][1], where `index` is a
+--     primary index or index that specified with argument `options.index`:
 --
 -- ```
--- local function iterate_with()
---     return true
--- end
+--  index:pairs(option.start_key(), {
+--     iterator = task.iterator_type
+--  }):take_while(
+--         function()
+--             return option.process_while()
+--         end
+--    )
 -- ```
+--
+--     [1]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_index/pairs/.
+--
```

> And I don't see any information about arguments, which is passed to the function.

it is a question how to describe it, because `task` is an internal object. Perhaps we need to change arguments passed to `iterate_with` to make possible to describe it in documentation. What do you think?

expirationd.lua
ligurio: Fixed.
```lua
@@ -513,7 +517,7 @@ end
 -- ```
 -- local function put_tuple_to_archive(space_id, args, tuple)
 --     box.space[space_id]:delete{tuple[1]}
---     local email = get_field(tuple, 2)
+--     local email = tuple[2]
 --     if args.archive_space_id ~= nil and email ~= nil then
 --         box.space[args.archive_space_id]:replace{email, fiber.time()}
 --     end
```


It was copy-pasted from TAP tests and `get_field` handle cases when tuple is `nil` field_no is bigger than `#tuple`:
```lua
-- get field                                                                     
local function get_field(tuple, field_no)                                        
    if tuple == nil then                                                         
        return nil                                                               
    end                                                                          
                                                                                 
    if #tuple < field_no then                                                    
        return nil                                                               
    end                                                                          
                                                                                 
    return tuple[field_no]                                                       
end                 
```

We don't need get_field< so I removed it.

expirationd.lua
ligurio: I keep it as is because I don't get how to fix it in LDoc.

expirationd.lua
ligurio: Fixed

Screenshot from 2021-09-27 17-11-07

expirationd.lua
Totktonada: It's good, but I meant a note somewhere at the module level. Maybe as part of usage example. If one don't know that there is a trick around `force` option to make expirationd work on a master with read only upstreams, (s)he will not know where to find the trick. It is the key idea of #14 (the issue appears despite the fact that the force option was mentioned in the readme).

expirationd.lua
Totktonada: Okay.

expirationd.lua
Totktonada: Okay, thanks! (I would place it somewhere in the documentation, but README is okay for me too.)

PR #73, opened 2021-09-09
Check types of function arguments with checks module

### LINKED ISSUE #58: Simplify source code with checking function arguments
In many places in `expirationd.lua` there are checks like below:
```lua
    if options.vinyl_assumed_space_len ~= nil then                               
        if type(options.vinyl_assumed_space_len) ~= 'number' then                
            error("Invalid type of vinyl_assumed_space_len value")               
        end                                                                      
        task.vinyl_assumed_space_len = options.vinyl_assumed_space_len           
    end  
```
it can be simplified with [checks](https://github.com/tarantool/checks) module that checks types of function arguments:
```lua
local options_types = {
    ...
    'number', -- options.vinyl_assumed_space_len
    ...
}

local function expirationd_run_task(name, space_id, is_tuple_expired, options)
checks('string', 'number', 'function', option_types)
task.vinyl_assumed_space_len = options.vinyl_assumed_space_len  
```

ligurio:
Closes #58

### DIFF
--- a/.github/workflows/fast_testing.yaml
+++ b/.github/workflows/fast_testing.yaml
@@ -40,12 +40,13 @@ jobs:
         id: cache-rocks
         with:
           path: .rocks/
-          key: cache-rocks-${{ matrix.tarantool }}-02
+          key: cache-rocks-${{ matrix.tarantool }}-04
 
       - name: Install requirements
         run: |
           tarantoolctl rocks install luatest 0.5.0
           tarantoolctl rocks install luacheck 0.26.0
+          tarantoolctl rocks make
         if: steps.cache-rocks.outputs.cache-hit != 'true'
 
       - run: make check
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -14,6 +14,7 @@
 -- local support functions
 -- ========================================================================= --
 
+local checks = require("checks")
 local fun = require("fun")
 local log = require("log")
 local fiber = require("fiber")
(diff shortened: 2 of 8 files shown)


expirationd.lua
```diff
@@ -580,6 +541,8 @@ end
 -- params:
 --   name -- task's name
 local function expirationd_task_stats(name)
+    checks('string')
```
olegrok, line 544: name is an optional argument. See code below. If name is nil then stats for all tasks is returned.

expirationd.lua
```diff
@@ -353,9 +354,27 @@ end
...
+        start_key = '?',
+        tuples_per_iteration = '?number',
+        atomic_iteration = '?boolean',
+        process_while = '?function',
+        iterate_with = '?function',
+        on_full_scan_start = '?function',
+        on_full_scan_complete = '?function',
+        on_full_scan_success = '?function',
+        on_full_scan_error = '?function',
+        args = '?table',
```
olegrok, line 370: It's not a table it's any

expirationd.lua
```diff
@@ -353,9 +354,27 @@ end
 --     * force                 -- Run task even on replica.
 --  }
 local function expirationd_run_task(name, space_id, is_tuple_expired, options)
-    if name == nil then
-        error("task name is nil")
-    end
+    checks('string', 'number|string', 'function', {
+        process_expired_tuple = '?function',
+        index = '?number|string',
+        iterator_type = '?string',
```
olegrok, line 360: Could it be box.index that it's actually number?

E.g.
```
tarantool> box.index.GE
---
- 5
...
```

expirationd.lua
```diff
@@ -353,9 +354,27 @@ end
...
+        process_while = '?function',
+        iterate_with = '?function',
+        on_full_scan_start = '?function',
+        on_full_scan_complete = '?function',
+        on_full_scan_success = '?function',
+        on_full_scan_error = '?function',
+        args = '?table',
+        full_scan_time = '?number',
+        iteration_delay = '?number',
+        full_scan_delay = '?number',
```
olegrok, line 373: I don't know is it relevant or not but maybe FFI numbers should be also available for numeric fields.

```
tarantool> type(1ULL)
---
- cdata
...

tarantool> type(1LL)
---
- cdata
...

tarantool> 1ULL == 1
---
- true
...
```

ligurio: Good point.

Added `cdata` for these arguments in `expirationd.start()`: full_scan_time, iteration_delay, full_scan_delay, tuples_per_iteration, vinyl_assumed_space_len, vinyl_assumed_space_len_factor.

expirationd.lua
ligurio: make it optional

.github/workflows/fast_testing.yaml
```diff
@@ -40,12 +40,13 @@ jobs:
...
         with:
           path: .rocks/
-          key: cache-rocks-${{ matrix.tarantool }}-02
+          key: cache-rocks-${{ matrix.tarantool }}-04
 
       - name: Install requirements
         run: |
           tarantoolctl rocks install luatest 0.5.0
           tarantoolctl rocks install luacheck 0.26.0
+          tarantoolctl rocks install checks 3.1.0
```
olegrok, line 49: I guess it should be `tarantoolctl rocks make` that download all needed dependencies

ligurio: Fixed for line with `checks`, other dependencies are not present in Luarock spec.

olegrok (approved):

expirationd.lua
Totktonada: NB: There are [checks](https://github.com/tarantool/checks/blob/eb49c159ac1f81d37420454c6574cc003aa8569d/checks.lua#L194-L227) for cdata numbers.

PR #74, opened 2021-09-09
Use actual methods names in regression tests

ligurio:


olegrok (approved):
This methods were renamed **six** years ago (41fb516f8ef38426bad59319a6bfe8020f5f0d37). Let's just drop them. They even aren't mentioned in readme.

ligurio:
> This methods were renamed six years ago (41fb516). Let's just drop them. They even aren't mentioned in readme.

obsoleted methods were not described in release notes, there were no messages about deprecation old methods.
We can broke someone if we drop these methods right now. I suggest to show messages and tell about it in release notes for next release. Then drop these old methods after 2-3 releases.

olegrok:
If anyone uses old option for six years we won't help him in 2-3 releases :).
Ok. I don't have any objections about merging this patch. But we probably should create an issue to remove this methods in future.

Side note: netbox:timeout() method was dropped after only 4 years of deprecation (https://github.com/tarantool/tarantool/commit/ad82655ebb590ee1a4ab077e3584227b5778a55e)

PR #75, opened 2021-09-09
test: add tests for expirationd.stats()

ligurio:


test/unit/stats_test.lua
```diff
@@ -0,0 +1,24 @@
...
+end)
+
+g.after_each(function()
+    helpers.truncate_spaces(g)
+end)
+
+function g.test_stats_simple()
+    expirationd.start("clean_all", g.tree.id, helpers.is_expired_true)
+    local stats = expirationd.stats("clean_all")
+    t.assert_equals(stats, {
```
olegrok, line 18: Don't you want to test each available counter? Currently this test only checks that `expirationd.stats` method just exists.

ligurio: Sure, will do.

test/unit/stats_test.lua
```diff
@@ -0,0 +1,79 @@
...
+
+g.before_all(function()
+    helpers.init_spaces(g)
+end)
+
+g.after_each(function()
+    helpers.truncate_spaces(g)
+end)
+
+function g.test_stats_basic()
```
olegrok, line 16: I think a general approach to such metrics should be following:
- Check that counter is 0
- Perform some actions
- Check that counter increased.

Currently there is an issue. Stats could have initially expected value and then test doesn't perform any action and finally we have expected result.

test/unit/stats_test.lua
```diff
@@ -0,0 +1,79 @@
...
+    local task = expirationd.start("stats_working_time", g.tree.id, helpers.is_expired_true)
+    local running_time = 1
+    fiber.sleep(running_time)
+
+    local stats = expirationd.stats("stats_working_time")
+    t.assert_equals(stats, {
+        checked_count = 0,
+        expired_count = 0,
+        restarts = 1,
+        working_time = running_time,
```
olegrok, line 76: I think this test could be flaky. In case of some slowdowns working_time could be greater than running_time.

test/unit/stats_test.lua
ligurio: I check initial counter's values in a first testcase. I don't see any reasons to check this in each testcase.

test/unit/stats_test.lua
ligurio: I believe patch below fixes it:

```lua
--- a/test/unit/expirationd_stats_test.lua
+++ b/test/unit/expirationd_stats_test.lua
@@ -66,14 +66,20 @@ end
 function g.test_stats_working_time()
     local task = expirationd.start("stats_working_time", g.tree.id, helpers.is_expired_true)
     local running_time = 1
+    local threshold = 0.5
+
+    local start_time = fiber.time()
     fiber.sleep(running_time)
+    local duration = fiber.time() - start_time
 
     local stats = expirationd.stats("stats_working_time")
+    t.assert_gt(stats.working_time, 0)
+    t.assert_lt(duration - stats.working_time, threshold)
+    stats.working_time = nil
     t.assert_equals(stats, {
         checked_count = 0,
         expired_count = 0,
         restarts = 1,
-        working_time = running_time,
     })
     task:kill()
 end
```

test/unit/expirationd_stats_test.lua
```diff
@@ -0,0 +1,84 @@
...
+    })
+    task:kill()
+end
+
+function g.test_stats_working_time()
+    local task = expirationd.start("stats_working_time", g.tree.id, helpers.is_expired_true)
+    local running_time = 1
+    local threshold = 0.3
+
+    local start_time = fiber.time()
```
olegrok, line 71: It's better to use fiber.clock for timeouts. fiber.time isn't monotonic

ligurio: it's not a timeout, it's an attempt to measure a rough operating time of task.

>  fiber.time isn't monotonic

overlooked it in documentation, thanks

olegrok (approved):

PR #81, opened 2021-09-22
Fix TAP regression tests for vinyl engine

ligurio:


Makefile
```diff
@@ -1,7 +1,8 @@
 CLEANUP_FILES  = tarantool.log
-CLEANUP_FILES += *.xlog*
+CLEANUP_FILES += *.xlog
+CLEANUP_FILES += *.vylog
```
olegrok, line 3: To be honest I'm not sure that it's really needed. AFAIU all vylog files are placed in 51* folders. Also vinyl produces run and index files but they are not mentioned here.

test.lua
```diff
@@ -142,14 +142,6 @@ local function init_box()
...
     local index_type = arg[1] or os.getenv('INDEX_TYPE') or 'TREE'
     local space_type = arg[2] or os.getenv('SPACE_TYPE') or 'memtx'
-    if space_type == 'vinyl' then
-        space_type = nil
-        if box.error.VINYL ~= nil then
-            space_type = 'vinyl'
-        else
-            os.exit(0)
-        end
-    end
```
olegrok, line 152: Please extend a commit message with an info about this change

test.lua
```diff
@@ -150,7 +150,7 @@ local function init_box()
     })
     a:create_index('first', {
         type = index_type,
-        parts = {1, 'NUM'},
+        parts = {1, 'unsigned'},
```
olegrok, line 153: nit: I think 'NUM' should be replaced with 'number' type. Unsigned is a subset of number.

olegrok: Also please fix count of brackets here: `parts = {1, type}` -> `parts = {{1, 'type'}}`. Or even `{{field = 1, type = type}}`.

test.lua
```diff
@@ -301,10 +301,7 @@ test:test('simple expires test',  function(test)
     local task = expirationd.task("test")
     test:is(task.start_time, start_time, 'checking start time')
     test:is(task.name, "test", 'checking task name')
-    local restarts = 5
-    if box.space[space_id].engine == 'memtx' then
-        restarts = 1
-    end
```
olegrok, line 315: Please describe your changes in commit message. It's not clear at first.

test.lua
```diff
@@ -432,7 +429,7 @@ test:test("multiple expires test", function(test)
     expirationd.start(
         "test",
         space_name,
-        check_tuple_expire_by_timestamp,
+        function() return true end,
```
olegrok, line 432: Why do you need this change?

test.lua
```diff
@@ -447,7 +444,7 @@ test:test("multiple expires test", function(test)
     fiber.sleep(1 + expire_delta)
     local task = expirationd.task("test")
     local cnt = task.expired_tuples_count
-    test:ok(cnt < tuples_count and cnt > 0, 'First part expires done')
+    test:ok(cnt <= tuples_count and cnt > 0, 'First part expires done')
```
olegrok, line 447: Why did it pass before?

test.lua
Totktonada: https://github.com/tarantool/tarantool/blob/9f208f0d3dc32ac3941090670c23d95f230ebd1a/src/box/field_def.c#L176-L191

sergepetrenko: As I can see now on customers trying to upgrade from 1.6 (where 'num' was still allowed) to 1.10, 'num' was really an alias for 'number'.
Even though our upgrade script changes 'num' fields to 'unsigned' (which was intended. See tarantool/tarantool#1534), users may have float/double/ whatever types in former 'num' fields. This leads to errors on upgrade.

I feel like 'num' should be replaced with 'unsigned' in our system spaces only.
And for all the other spaces (user defined/testing) 'number' is a better substitution to 'num'.

test.lua
ligurio: Done.

test.lua
ligurio: Testcase has been fixed in another way.

test.lua
```diff
@@ -16,6 +17,54 @@ strict.on()
...
+    --
+    -- cond_function_res_1 is the 'ok' / 'not ok' indicator for
+    -- wait_cond(). But we return it: this is deliberate choice.
+    -- This way a user may understand, whether we finally hit
+    -- timeout or not.
+    if not res[1] then
+        error(res[2])
+    end
+    return select(2, unpack(res))
+end
```
olegrok, line 36: I guess it should be done in the way as box.atomic work (https://github.com/tarantool/tarantool/blob/298eebf7c8e8dc2c985f20e22d35ef5c1bfd7c17/src/box/lua/schema.lua#L356)

test.lua
```diff
@@ -16,6 +17,54 @@ strict.on()
...
+    --      ...
+    --
+    -- cond_function_res_1 is the 'ok' / 'not ok' indicator for
+    -- wait_cond(). But we return it: this is deliberate choice.
+    -- This way a user may understand, whether we finally hit
+    -- timeout or not.
+    if not res[1] then
+        error(res[2])
+    end
+    return select(2, unpack(res))
```
olegrok, line 35: unpack for {1, nil, 3} returns only 1. Should be replaced to unpack(res, 1, table.maxn(res)).

test.lua
```diff
@@ -452,15 +493,26 @@ test:test("multiple expires test", function(test)
...
     )
     -- test first expire part
-    fiber.sleep(1 + expire_delta)
-    local task = expirationd.task("test")
-    local cnt = task.expired_tuples_count
-    test:ok(cnt < tuples_count and cnt > 0, 'First part expires done')
+    local res = wait_cond(
+        function()
+            local task = expirationd.task("test")
+            local cnt = task.expired_tuples_count
```
olegrok, line 499: Shouldn't we compare this value with actual records count in space? (In similar way it does here - https://github.com/tarantool/expirationd/pull/81/files#diff-8cb3361cd200c46658a6738e92c421b0589c52742f6160315b03bc4192a2951fR527)

test.lua
ligurio: Good point. Fixed.

test.lua
ligurio: If I get it right you want a rollback in case of cond function fails.
However everywhere wait_cond() is used for polling stats from expirationd and wait_cond() doesn't execute any operations with data. So using something like `box.atomic()` is not applicable here.

PR #83, opened 2021-09-24
Take maintainership

ligurio:


expirationd-scm-1.rockspec
```diff
@@ -8,7 +8,7 @@ description = {
     summary = "Expiration daemon for Tarantool",
     homepage = "https://github.com/tarantool/expirationd",
     license = "BSD2",
-    maintainer = "Eugine Blikh <bigbes@tarantool.org>"
+    maintainer = "Sergey Bronnikov <sergeyb@tarantool.org>"
```
Totktonada, line 11: There is a maintainer field in `debian/control` as well.

ligurio: Updated too.

Totktonada (approved):
LGTM.

With great power comes great responsibility!

PR #87, opened 2021-10-15
Fix iterate_with function in tests

### LINKED ISSUE #84: Wrong iterate_with function in tests
Tests for `iterate_with` uses wrong implementation of `iterate_with` function. It must return an iterator but it always returns `true`:

test/helper.lua
```lua
function helpers.is_expired_true()
    return true
end
```

ligurio:
Closes #84


test/helper.lua
```diff
@@ -99,8 +99,13 @@ function helpers.truncate_spaces(g)
     g.vinyl:truncate()
 end
 
-function helpers.is_expired_true()
-    return true
+function helpers.is_expired_true(task)
```
olegrok, line 102: In description you say "iterate_with" but here you change is_expired_true. It seems you fixed wrong function.

ligurio: Right, overlooked that it.

PR #88, opened 2021-10-18
ci: add reusable testing workflow

ylobankov:
The idea of this workflow is to be a part of the 'integration.yml'
workflow from the tarantool repo to verify the integration of the
expirationd module with an arbitrary tarantool version.

This workflow is not triggered on a push to the repo and cannot be run
manually since it has only the 'workflow_call' trigger. This workflow
will be included in the tarantool development cycle and called by the
'integration.yml' workflow from the tarantool project on a push to the
master and release branches for verifying integration of tarantool with
expirationd.

Part of tarantool/tarantool#6528
Part of tarantool/tarantool#5265
Part of tarantool/tarantool#6056

Workflow on Tarantool side: https://github.com/tarantool/tarantool/pull/6529

### DIFF
--- a/.github/workflows/reusable_testing.yml
+++ b/.github/workflows/reusable_testing.yml
@@ -1,0 +1,33 @@
+name: reusable_testing
+
+on:
+  workflow_call:
+    inputs:
+      artifact_name:
+        description: The name of the Tarantool build artifact
+        default: ubuntu-focal
+        required: false
+        type: string
+
+jobs:
+  run_tests:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Clone the expirationd module
+        uses: actions/checkout@v2
+        with:
+          repository: ${{ github.repository_owner }}/expirationd
+
+      - name: Download the Tarantool build artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: ${{ inputs.artifact_name }}
+
+      - name: Install Tarantool
+        # Now we're lucky: all dependencies are already installed. Check package
+        # dependencies when migrating to other OS version.
+        run: sudo dpkg -i tarantool_*.deb tarantool-common_*.deb tarantool-dev_*.deb
+
+      - run: make deps
+      - run: echo $PWD/.rocks/bin >> $GITHUB_PATH
+      - run: make test


ylobankov:
Related PR https://github.com/tarantool/tarantool/pull/6529

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,52 @@
...
+        with:
+          name: ${{ inputs.artifact_name }}
+      # TODO(ylobankov): Just in case remove tarantool that could be previously
+      # installed on the runner. Relevant to self-hosted runners. Remove this
+      # step when only GitHub-hosted runners are used for integration testing.
+      - run: sudo apt-get -y purge 'tarantool*'
+      - name: 'Install tarantool'
+        # TODO(ylobankov): Install package dependencies. Now we're lucky: all
+        # dependencies are already there.
+        run: sudo dpkg -i tarantool*.deb
```
ligurio, line 32: I don't like `*` as a mask, let's specify what exactly we want to install. For example: `dpkg -i tarantool tarantool-dev tarantool-common`.

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: 'Install expirationd'
```
ligurio, line 34: Could you specify in name that we install 'current' version of expirationd? `Install expirationd -current`

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,52 @@
...
+        run: tarantoolctl rocks make
+
+      - name: 'Install test requirements'
+        run: tarantoolctl rocks install luatest 0.5.6
+
+      - name: 'Run tests'
+        run: .rocks/bin/luatest -v -b
```
ligurio, line 41: `make test`

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      # TODO(ylobankov): Relevant to self-hosted runners. Remove this step
+      # when only GitHub-hosted runners are used for integration testing.
+      - if: always()
+        run: rm -rf .rocks
+
+      # TODO(ylobankov): Relevant to self-hosted runners. Remove this step
+      # when only GitHub-hosted runners are used for integration testing.
+      - if: always()
+        run: sudo apt-get -y purge 'tarantool*'
```
ligurio, line 51: The same as above - let's specify package names instead of mask.

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,52 @@
...
+    steps:
+      - name: 'Clone the expirationd module'
+        uses: actions/checkout@v2
+        with:
+          repository: ${{ github.repository_owner }}/expirationd
+
+      - name: 'Download the tarantool build artifact'
+        uses: actions/download-artifact@v2
+        with:
+          name: ${{ inputs.artifact_name }}
```
ligurio, line 24: As I get it right we will download each Tarantool build via network. Network is unreliable and I suspect there is possibility of flaky runs due to network problems. Is it possible to setup download with retries and timeouts?

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,52 @@
+name: reusable_testing
+
+on:
+  workflow_call:
+    inputs:
+      artifact_name:
+        description: 'The name of the tarantool build artifact'
```
ligurio, line 7: tarantool -> Tarantool, please

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,52 @@
...
+jobs:
+  run_tests:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: 'Clone the expirationd module'
+        uses: actions/checkout@v2
+        with:
+          repository: ${{ github.repository_owner }}/expirationd
+
+      - name: 'Download the tarantool build artifact'
```
ligurio, line 21: tarantool -> Tarantool, please

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,52 @@
...
+        uses: actions/download-artifact@v2
+        with:
+          name: ${{ inputs.artifact_name }}
+      # TODO(ylobankov): Just in case remove tarantool that could be previously
+      # installed on the runner. Relevant to self-hosted runners. Remove this
+      # step when only GitHub-hosted runners are used for integration testing.
+      - run: sudo apt-get -y purge 'tarantool*'
+      - name: 'Install tarantool'
```
ligurio, line 29: tarantool -> Tarantool, please

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,52 @@
...
+        # TODO(ylobankov): Install package dependencies. Now we're lucky: all
+        # dependencies are already there.
+        run: sudo dpkg -i tarantool*.deb
+
+      - name: 'Install expirationd'
+        run: tarantoolctl rocks make
+
+      - name: 'Install test requirements'
+        run: tarantoolctl rocks install luatest 0.5.6
```
ligurio, line 38: Please make a separate target in Makefile for requirements installation (see https://github.com/tarantool/expirationd/blob/master/.github/workflows/fast_testing.yaml#L49-L56) and let's use `make deps` in workflows.

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,52 @@
+name: reusable_testing
```
ligurio, line 1: Why 'reusable' if it is an integration testing?

.github/workflows/reusable_testing.yml
ylobankov: Well, I need the only test requirement here - luatest. I don't think that we need to install luacheck, luacov, luacov-coveralls for such kind of testing that is done here. IMHO, It is redundant. So I decided to install exactly what I need. But for the fast_testing.yml workflow, creating a separate target for requirements installation will make sense.

.github/workflows/reusable_testing.yml
ylobankov: The same, `make test` enables coverage that is not important for us. That is why I just execute the command that runs tests. Actually, I can create such targets in Makefile that I need here, but I don't see any profit in that. Maybe I miss something :)

.github/workflows/reusable_testing.yml
Totktonada: Sure, it is more like a template with 'build expirationd and run against given tarantool artifact' steps rather than a real workflow. We agreed on this name with Yaroslav during work on https://github.com/tarantool/vshard/pull/299. I had a wish to have something like 'reusable' or 'callable' in the name and we stick on the former, because GitHub's feature itself is called 'reusable workflows'.

We agreed on the 'integration' name for the caller workflow in tarantool (because it is about integration testing and because 'integration_testing' is too long for GitHub UI).

.github/workflows/reusable_testing.yml
Totktonada: It is the file glob: install what is extracted from the artifact.

.github/workflows/reusable_testing.yml
ylobankov: Probably if we chose `callable` instead of `reusable`, maybe it would confuse people less. What do you think?

.github/workflows/reusable_testing.yml
ylobankov: I hope this will disappear in this PR due to https://github.com/tarantool/tarantool/pull/6537.

ligurio:
Added to the PR description a link to PR for workflow on Tarantool side -https://github.com/tarantool/tarantool/pull/6529.

.github/workflows/reusable_testing.yml
ylobankov: Done.

.github/workflows/reusable_testing.yml
ylobankov: Not relevant now.

.github/workflows/reusable_testing.yml
ylobankov: Done.

.github/workflows/reusable_testing.yml
ylobankov: Not relevant now.

.github/workflows/reusable_testing.yml
ylobankov: Done.

ligurio:
> and now this PR blocked by #89 as in #89 we added make deps used here.

#89 is merged

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          repository: ${{ github.repository_owner }}/expirationd
+
+      - name: 'Download the Tarantool build artifact'
+        uses: actions/download-artifact@v2
+        with:
+          name: ${{ inputs.artifact_name }}
+
+      - name: 'Install Tarantool'
+        # TODO(ylobankov): Install package dependencies. Now we're lucky: all
```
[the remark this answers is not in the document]
Totktonada: If we'll add a runtime dependency into tarantool and it appears that it is not installed by default on GitHub virtual runners, we'll hit by the problem here (and in other modules): `dpkg -i file` does not install dependencies. So I would consider it as problematic code that should be fixed sooner or later.

It is easy for CentOS / Fedora (just use `yum install -y file` instead of `rpm -i file`). We should find something like this for Debian / Ubuntu (maybe use apt? whether it is available on Debian or it is pure Ubuntu tool?).

The question was raised first time [here](https://github.com/tarantool/smtp/blob/ed6434629cd0ab4514de13e5e1780a03903f403b/.github/workflows/packaging.yml#L146-L166).

Anyway, it is up to @ligurio how to better present this information here. I just share my opinion.

.github/workflows/reusable_testing.yml
ylobankov: @ligurio Should we consider this discussion resolved?

ligurio: > It is the file glob: install what is extracted from the artifact.

I don't know what was extracted from the artifact.
Why we cannot run dpkg with package name and glob instead of version part?

ylobankov: You mean `sudo dpkg -i tarantool_*.deb tarantool-common_* tarantool-dev_*`?

Totktonada: I'm not against explicit package names (explicit is better than implicit, sure), but I would highlight my personal opinion about the tarantool-command package: I think it [should be removed](https://github.com/tarantool/tarantool/issues/6400) in a future.

`dpkg -i tarantool*.deb` will continue to work on all tarantool versions, but `dpkg -i tarantool_*.deb tarantool-common_*.deb tarantool-dev_*.deb` should be transformed into something like the following (after change in tarantool):

```sh
# gh-6400: blah blah.
if [ -f tarantool-common_*.deb ]; then
    dpkg -i tarantool-common_*.deb
fi
dpkg -i tarantool_*.deb tarantool-dev_*.deb
```

But I'll repeat that the linked issue reflects only my personal opinion (however I didn't heard the opposite opinions) and there are no certain plans to actually implement it.

ligurio: > You mean sudo dpkg -i tarantool_*.deb tarantool-common_* tarantool-dev_*?

Exactly.

ylobankov: Ok, I don't mind. Fixed.

.github/workflows/reusable_testing.yml
ligurio: Okay, let's keep the name of workflow as is. Anyway, we already use the same name in other external repositories, and it would be good to make it align with them.

ligurio:
> Seems we have flaky tests in expirationd (#90) and we need to fix them before merging this PR.

Fix is on review https://github.com/tarantool/expirationd/pull/97.

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,35 @@
...
+        uses: actions/download-artifact@v2
+        with:
+          name: ${{ inputs.artifact_name }}
+
+      - name: Install Tarantool
+        # Now we're lucky: all dependencies are already installed. Check package
+        # dependencies when migrating to other OS version.
+        run: >
+          sudo dpkg -i tarantool_*.deb tarantool-common_*.deb
+          tarantool-dev_*.deb
```
ligurio, line 31: Why the list is split for two lines?

ligurio (changes requested):

.github/workflows/reusable_testing.yml
ylobankov: For more readability. I try to keep the length of the line equal to 80 characters.

ligurio: If so, then we need to fix the indentation for better readability :)

```
          sudo dpkg -i tarantool_*.deb tarantool-common_*.deb
                       tarantool-dev_*.deb
```

ylobankov: May be
```
sudo dpkg -i tarantool_*.deb tarantool-common_*.deb 
    tarantool-dev_*.deb
```
?

ligurio: Package names should start at the same position on both lines.

ylobankov: Fixed.

VitaliyaIoffe (approved):

Issue #92, opened 2021-11-03
Incorrect description of expiration for replicas


ArtDu:
> Note about using expirationd with replication: by default expirationd should process tasks only on the writeable instance, it means that expirationd will not start task processing on a replica. Here the word 'replica' means an instance with at least one configured upstream, it's an option box.cfg.replication_source (box.cfg.replication for Tarantool 1.7.6+)

The expiration seems to behave differently. It will not perform expiration even if the expiration is started on the master. Because if the master is in the replica set, then box.cfg.replication will not be nil. It should be added that expiration will not occur if the instance is included in the replica set, and not specifically on the replica or master.

https://github.com/tarantool/expirationd/blob/513fe16caf31d69882a56b4047cc449927cb3074/expirationd.lua#L151

Totktonada:
Looks connected to #96.

PR #94, opened 2021-11-03
Update README: installation steps, prerequisites and dependencies

### LINKED ISSUE #93: Documentation enhancement
I think it is worth clarifying the description of the installation process by mentioning that you need to install dependencies by adding `make deps` string.

ligurio:
Closes #93

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,10 +59,9 @@ Note about using expirationd with replication: by default expirationd should
 process tasks only on the writeable instance, it means that expirationd *will
 not* start task processing on a replica. Here the word 'replica' means an
 instance with at least one configured upstream, it's an option
-`box.cfg.replication_source` (`box.cfg.replication` for Tarantool 1.7.6+). One
-can force running task on replica with option `force` in `start()` module
-function. The option force let a user control where to start task processing
-and where don't.
+`box.cfg.replication_source`. One can force running task on replica with option
+`force` in `start()` module function. The option force let a user control where
+to start task processing and where don't.
 
 ### Examples
 
--- a/README.md
+++ b/README.md
@@ -111,4 +111,7 @@ expirationd.start(job_name, space.id, is_expired, {
 
 ## Testing
 
-Simply start `make test`
+```
+$ make deps
+$ make test
+```
--- a/README.md
+++ b/README.md
@@ -31,6 +31,26 @@ Table below may help you to choose a proper module for your requirements:
 | expirationd   | Medium (sec)  | Yes          | Yes                 | synchronous (fiber with condition) |
 | moonwalker    | NA            | No           | Yes                 | asynchronous (using crontab etc)   |
 
+### Prerequisites
+
+* Tarantool 1.10+ (`tarantool` package, see [documentation](https://www.tarantool.io/en/download/)).
+
+### Installation
+
+You can:
+
+* Install the module using `tarantoolctl`:
+
+  ``` bash
+  tarantoolctl rocks install expirationd
+  ```
+
+* Install the module using LuaRocks:
+
+  ``` bash
+  luarocks install --local --server=https://rocks.tarantool.org expirationd
+  ```
+
 ### Documentation
 
 See API documentation in https://tarantool.github.io/expirationd/


README.md
```diff
@@ -31,6 +31,26 @@ Table below may help you to choose a proper module for your requirements:
 | expirationd   | Medium (sec)  | Yes          | Yes                 | synchronous (fiber with condition) |
 | moonwalker    | NA            | No           | Yes                 | asynchronous (using crontab etc)   |
 
+### Prerequisites
+
+* Tarantool 1.10+ (`tarantool` package, see [documentation](https://www.tarantool.io/en/download/)).
```
olegrok, line 36: It's a bit inconsistent with phrase above:
```
it's an option box.cfg.replication_source (box.cfg.replication for Tarantool 1.7.6+)
```

So... You need to drop 1.10- mentions from readme to use 1.10+ as prerequisite

ligurio: Hmm... I don't want to declare work with 1.7.6 because we don't test it on CI.
However, 1.7.6 should work too.

olegrok: If you can't guarantee 1.7.6, let's remove it from readme.

ligurio: removed

README.md
```diff
@@ -39,7 +59,7 @@ Note about using expirationd with replication: by default expirationd should
 process tasks only on the writeable instance, it means that expirationd *will
 not* start task processing on a replica. Here the word 'replica' means an
 instance with at least one configured upstream, it's an option
-`box.cfg.replication_source` (`box.cfg.replication` for Tarantool 1.7.6+). One
+`box.cfg.replication_source` (`box.cfg.replication` for Tarantool 1.10+). One
```
olegrok, line 62: Let's also drop info about`box.cfg.replication_source`. 1.10 shouldn't be in brackets as some kind of clarification.

ligurio: Why not?

olegrok: Modern users don't know anything about box.cfg.replication_source. Or at least it should be changed to "box.cfg.replication (box.cfg.replication_source for Tarantool before 1.7.6)". This text was written when majority of users used 1.6. Currently it's not so.

ligurio: This text was written month ago in commit fbdf9965.

olegrok: Well, but it's anyway quite strange to mention box.cfg.replication_source as "primary" when it was deprecated (or even removed four years ago https://github.com/tarantool/tarantool/commit/c386c1932c8087a79a523792e3cfcc8b960855fd)

Totktonada: It was not deleted and still kept (AFAIR, gives a warning about using of a deprecated option).

I agree, there is no sense to declare support of 1.10+ and tell about the option that was removed in the 1.7 era. At least not in the main text (if there are footnotes support, we can add something like 'box.cfg.replication_source is accepted too, but may be removed in a future, see [link to docs]').

ligurio: I think we don't need to say anything special about ``. Everything works without actions from user side.
```lua
        if (box.cfg.replication_source == nil and box.cfg.replication == nil) or task.force then
```

Updated a patch:
```
--- a/README.md
+++ b/README.md
@@ -59,10 +59,9 @@ Note about using expirationd with replication: by default expirationd should
 process tasks only on the writeable instance, it means that expirationd *will
 not* start task processing on a replica. Here the word 'replica' means an
 instance with at least one configured upstream, it's an option
-`box.cfg.replication_source` (`box.cfg.replication` for Tarantool 1.10+). One
-can force running task on replica with option `force` in `start()` module
-function. The option force let a user control where to start task processing
-and where don't.
+`box.cfg.replication_source`. One can force running task on replica with option
+`force` in `start()` module function. The option force let a user control where
+to start task processing and where don't.
```

README.md
```diff
@@ -59,10 +59,9 @@ Note about using expirationd with replication: by default expirationd should
 process tasks only on the writeable instance, it means that expirationd *will
 not* start task processing on a replica. Here the word 'replica' means an
 instance with at least one configured upstream, it's an option
-`box.cfg.replication_source` (`box.cfg.replication` for Tarantool 1.7.6+). One
-can force running task on replica with option `force` in `start()` module
-function. The option force let a user control where to start task processing
-and where don't.
+`box.cfg.replication_source`. One can force running task on replica with option
+`force` in `start()` module function. The option force let a user control where
+to start task processing and where don't.
```
ArtDu, line 64: can you fix this(https://github.com/tarantool/expirationd/issues/92) ticket additionally, since you are just making changes about box.cfg.replication? Or is it better to do it in another PR?

ligurio: Sorry, no. It's better to create a separate PR.

olegrok (approved):
I don't understand a change of "replication_source" (should be just replication, because I don't know projects that use replication_source. Also there is a cartridge that automatically configures replication option) but OK, let's fix it later.

Totktonada:
> I don't understand a change of "replication_source" (should be just replication, because I don't know projects that use replication_source.

I'm too. Sergey, please, fix it when you'll touch the documentation next time.

PR #108, opened 2022-05-12
Prevent iteration through a functional index for Tarantool < 2.8.4

### LINKED ISSUE #101: Functional index iterator is not stable
So we can crash tarantool and observe other unwanted effects. Consider https://github.com/tarantool/tarantool/issues/6786 for details.

See https://github.com/tarantool/doc/issues/2102 and further links for some kind of definition of the 'stable iterator' term. Say, you open an iterator by a space index, moving on it and changing something in the space. The iterator is 'stable' when it is not invalidated by the changes in the space. It always has a predictable behavior and don't lead to crash, data corruption or other undesirable effects.

The most common use case for a stable iterator is to delete tuples while iterating over a space:

```lua
for _, tuple in box.space.s.index.foo:pairs() do
    box.space.s:delete({tuple[1]})
end

```

However it is not the only use case.

The problem can be reproduced on a debug tarantool build. It reproduces for me on expirationd 1.1.1-44-g838c2d1 on the `test_memtx_tree_functional_index` test case.

```sh
$ cd /home/alex/projects/tarantool-meta/r/t-6
$ cmake . -DCMAKE_BUILD_TYPE=Debug -DENABLE_BACKTRACE=ON -DENABLE_DIST=ON -DENABLE_FEEDBACK_DAEMON=OFF -DENABLE_BUNDLED_LIBCURL=OFF && make -j
$ cd /home/alex/projects/tarantool-meta/expirationd
$ export PATH="/home/alex/projects/tarantool-meta/r/t-6/src:${PATH}"
$ luatest -c -v -p test_memtx_tree_functional_index
<...>
    custom_index.engine:"memtx".index_type:"TREE".test_memtx_tree_functional_index ... 
2022-01-15 01:59:22.159 [7804] main/110/guardian of "clean_all" I> expiration: task "clean_all" restarted
tarantool: /home/alex/p/tarantool-meta/r/t-6/src/box/tuple_compare.cc:1394: int func_index_compare(tuple*, hint_t, tuple*, hint_t, key_def*) [with bool is_nullable = false; hint_t = long unsigned int]: Assertion `mp_typeof(*key_b) == MP_ARRAY' failed.
Aborted (core dumped)
```

According to what I said [here](https://github.com/tarantool/expirationd/pull/40#issuecomment-825983154), a year ago we didn't lean on the tree iterator stability, so the test would likely pass. So I guess we can consider this bug as a regression. I didn't check, though.

It seems, we should switch to the old 'get batch, process, get next batch using GT iterator' iteration way for functional indexes.

It was not catched by tests before, because we have no test runs on a debug tarantool build in CI. It seems, it is must have for expirationd, where we lean on non-trivial index guarantees. I filed #102 to add testing on the debug build into CI.

oleg-jukovec:
expirationd.start fails with an error for a functional index if
a Tarantool instance version is less than 2.8.4. It can be disabled
using options.force_allow_functional_index.

Closes #101

An user can use options.force_broken + options.iterate_with to avoid the crash, see:
https://github.com/tarantool/expirationd/blob/a2bc3d39ea20197818170da1886adfdd3de4cfac/test/unit/custom_index_test.lua#L239-L245

### DIFF
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -450,7 +450,12 @@ end
 --     configured upstream, it's an option `box.cfg.replication_source`
 --     (`box.cfg.replication` for Tarantool 1.7.6+). The option `force` let a
 --     user control where to start task processing and where don't.
---
+-- @boolean[opt] options.force_allow_functional_index
+--     By default expirationd returns an error on iteration through a functional
+--     index for Tarantool < 2.8.4 because it may cause a crash, see
+--     https://github.com/tarantool/expirationd/issues/101
+--     You can skip the error using the option if you know what you are doing
+--     (implement your own `iterate_with` as example).
 -- @number[opt] options.full_scan_delay
 --     Sleep time between full scans (in seconds). It is allowed to pass an FFI
 --     number: `1LL`, `1ULL` etc. Default value is 1 sec.
@@ -603,6 +608,7 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
         args = '?',
         atomic_iteration = '?boolean',
         force = '?boolean',
+        force_allow_functional_index = '?boolean',
         full_scan_delay = '?number|cdata',
         full_scan_time = '?number|cdata',
         index = '?number|string',
@@ -649,6 +655,27 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
         expire_index = box.space[space_id].index[options.index]
         if expire_index.type ~= "TREE" and expire_index.type ~= "HASH" then
             error("Not supported index type, expected TREE or HASH")
+        end
+        local engine = box.space[space_id].engine
+        if engine == "memtx" and expire_index.func ~= nil then
+            local supported = false
+            local version = rawget(_G, "_TARANTOOL"):split('-', 1)[1]
+            local major_minor_patch = version:split('.', 2)
+
+            local major = tonumber(major_minor_patch[1])
+            local minor = tonumber(major_minor_patch[2])
+            local patch = tonumber(major_minor_patch[3])
+            -- https://github.com/tarantool/expirationd/issues/101
+            -- fixed since 2.8.4 and 2.10
+            if (major > 2) or (major == 2 and minor == 8 and patch >= 4)
+               or (major == 2 and minor >= 10) then
+                 supported = true
+            end
+            local force_allow = options.force_allow_functional_index or false
+            if not supported and not force_allow then
+                error("Functional indices are not supported for" ..
+                      " Tarantool < 2.8.4, see options.force_allow_functional_index")
+            end
         end
     end
     task.index = expire_index
--- a/test/unit/custom_index_test.lua
+++ b/test/unit/custom_index_test.lua
@@ -204,8 +204,63 @@ function g.test_tree_index_multikey(cg)
     task:kill()
 end
 
+function g.test_memtx_tree_functional_index_broken(cg)
+    t.skip_if(_TARANTOOL < "2" or not helpers.memtx_func_index_is_broken(),
+              "Unsupported Tarantool version")
+    t.skip_if(cg.params.index_type ~= 'TREE', 'Unsupported index type')
+    t.skip_if(cg.params.engine == 'vinyl', 'Unsupported engine')
+    local expected = "Functional indices are not supported for Tarantool < 2.8.4," ..
+                     " see options.force_allow_functional_index"
+
+    cg.space:insert({1, "1", nil, nil, nil, nil, nil, "12"})
+    cg.space:insert({2, "2", nil, nil, nil, nil, nil, "21"})
+
+    t.assert_error_msg_contains(expected, expirationd.start, "clean_all", cg.space.id,
+            helpers.is_expired_debug, {index = "functional_index"})
+    t.assert_items_equals(cg.space:select({}, {limit = 10}), {
+            {2, "2", nil, nil, nil, nil, nil, "21"},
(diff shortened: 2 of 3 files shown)


expirationd.lua
```diff
@@ -650,6 +656,28 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
...
+            local sdk = rawget(_G, "_TARANTOOL")
+            local major_minor_patch = sdk:split('-', 1)[1]
+            local major_minor_patch_parts = major_minor_patch:split('.', 2)
+
+            local major = tonumber(major_minor_patch_parts[1])
+            local minor = tonumber(major_minor_patch_parts[2])
+            local patch = tonumber(major_minor_patch_parts[3])
+            -- https://github.com/tarantool/expirationd/issues/101
+            -- fixed since 2.8.4 and 2.10
+            if (major == 2 and minor == 8 and patch >= 4)
```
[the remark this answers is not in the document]
oleg-jukovec: major > 2 is a separate case, please, see the line below:
```lua
or (major == 2 and minor >= 10) or (major > 2) then
```

expirationd.lua
```diff
@@ -650,6 +656,28 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
         if expire_index.type ~= "TREE" and expire_index.type ~= "HASH" then
             error("Not supported index type, expected TREE or HASH")
         end
+        local engine = box.space[space_id].engine
+        if engine == "memtx" and expire_index.func ~= nil then
+            local supported = false
+            local sdk = rawget(_G, "_TARANTOOL")
```
Totktonada, line 662: Nit: `sdk` is not appropriate term here IMHO. Just `version` would look better. Or `_G._TARANTOOL:split(<...>)`.

test/unit/custom_index_test.lua
```diff
@@ -204,8 +204,50 @@ function g.test_tree_index_multikey(cg)
...
+    t.skip_if(cg.params.index_type ~= 'TREE', 'Unsupported index type')
+    t.skip_if(cg.params.engine == 'vinyl', 'Unsupported engine')
+    local expected = "Functional indices are not supported for Tarantool < 2.8.4," ..
+                     " see options.force_broken"
+
+    t.assert_error_msg_contains(expected, expirationd.start, "clean_all", cg.space.id,
+            helpers.is_expired_debug, {index = "functional_index"})
+end
+
+function g.test_memtx_tree_functional_index_force_broken(cg)
```
Totktonada, line 219: I afraid this test may be unstable / crashing. Can we run the case on an empty space (just to check that the error is not raised)? I guess it should be safe.

expirationd.lua
```diff
@@ -603,6 +608,7 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
         args = '?',
         atomic_iteration = '?boolean',
         force = '?boolean',
+        force_broken = '?boolean',
```
Totktonada, line 611: I think that 'functional index' words should be in the option name. Say, `forcely_allow_functional_index`. A long name is okay for a dangerous option of the I-know-what-I-an-doing kind, IMHO.

test/unit/custom_index_test.lua
Totktonada: Ah, I see now: you select all tuples at once before all deletions. Good solution.

It would be good to mark this place with a short comment. Not obvious at first glance.

oleg-jukovec: Thank you, I rewrote the test: insert + expirationd.start + checks + truncate. Now it's obvious.

expirationd.lua
oleg-jukovec: No problems, I changed it to ```force_allow_functional_index```

expirationd.lua
```diff
@@ -450,7 +450,12 @@ end
 --     configured upstream, it's an option `box.cfg.replication_source`
```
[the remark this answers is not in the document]
oleg-jukovec: Thanks, I clicked the "ready to view".

test/unit/custom_index_test.lua
Totktonada: It seems you understood me wrongly. I'm about the 'force broken' test case: when the `force_allow_functional_index` is enabled. Especially, about this chunk of the code:

```lua
local select_with = function(task)
    return pairs(task.index:select({}, {iterator = "ALL"}))
end
```

I just want a short comment, why it is necessary.

Totktonada (approved):
LGTM.

test/unit/custom_index_test.lua
oleg-jukovec: Yes, sorry for the misunderstanding. I've added a comment:
```patch
--- a/test/unit/custom_index_test.lua
+++ b/test/unit/custom_index_test.lua
@@ -236,8 +236,11 @@ function g.test_memtx_tree_functional_index_force_broken(cg)
     cg.space:insert({1, "1", nil, nil, nil, nil, nil, "12"})
     cg.space:insert({2, "2", nil, nil, nil, nil, nil, "21"})
 
+    -- The problem occurs when we iterate through a functional index and delete
+    -- a current tuple. A possible solution is somehow to process tuples chunk
+    -- by chunk using select calls instead of iterating with index:pairs().
     local select_with = function(task)
-        return pairs(task.index:select({}, {iterator = "ALL"}))
+        return pairs(task.index:select({}, {iterator = "ALL", limit = 100}))
     end
```

Totktonada: Thanks for the update!

PR #115, opened 2022-06-24
ci: use setup-tarantool for 2.10

Totktonada:
Now the workaround for https://github.com/tarantool/setup-tarantool/issues/19 is not needed, because tarantool-2.10+ releases are supported since setup-tarantool v1.3.0.

The action uses the GitHub's caching infrastructure to speed up tarantool installation.


oleg-jukovec (approved):
Thank you for the patch!

PR #118, opened 2022-07-09
Shuffle tests

ligurio:
PR introduces two changes:

1. shuffling tests in luatest
    luatest has an option --shuffle that allows to randomise the order of
    tests. This can be useful to detect a test that passes just because it
    happens to run after an unrelated test that leaves the system in a
    favourable state. Patch enable tests shuffling with predefined random
    seed specified by $RANDOM. Seed is useful for reproducing problems
    related to the specific test order.
2. bump luatest version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ LUACOV_REPORT := $(PROJECT_DIR)/luacov.report.out
 LUACOV_STATS := $(PROJECT_DIR)/luacov.stats.out
 
 SHELL := $(shell which bash) # Required for brace expansion used in a clean target.
+SEED ?= $(shell /bin/bash -c "echo $$RANDOM")
 
 CLEANUP_FILES  = tarantool.log
 CLEANUP_FILES += *.index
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -7,6 +7,7 @@ LUACOV_REPORT := $(PROJECT_DIR)/luacov.report.out
 LUACOV_STATS := $(PROJECT_DIR)/luacov.stats.out
 
 SHELL := $(shell which bash) # Required for brace expansion used in a clean target.
+SEED := $(shell /bin/bash -c "echo $$RANDOM")
```
Totktonada, line 10: Maybe it worth to allow to redefine it externally: `SEED ?= <...>`.

Makefile
```diff
@@ -32,7 +33,7 @@ luacheck:
 
 .PHONY: test
 test:
-	luatest -v --coverage
+	luatest -v --coverage --shuffle all:${SEED} --seed ${SEED}
```
Totktonada, line 36: We can pass it once: https://github.com/tarantool/luatest/blob/f639ce4c75808166163c60b50e88a7a942b3b885/luatest/runner.lua#L137-L144.

oleg-jukovec (approved):
Thank you! LGTM.

Totktonada (approved):

PR #120, opened 2022-07-11
test: fix test_mvcc_vinyl_tx_conflict

### LINKED ISSUE #105: test: test_mvcc_vinyl_tx_conflict() permanently fails starting since Tarantool 2.10.0-beta2-169-gb9f6d3858
Environment: Ubuntu 20.04
Tarantool: 2.10.0-beta2-169-gb9f6d3858
Expirationd: 838c2d16346741cd753f6a95810284ecb4c6b50a (current master at this moment)

```
Failed tests:
-------------

1) atomic_iteration.engine:"vinyl".index_type:"TREE".test_mvcc_vinyl_tx_conflict
...orkspace/expirationd/test/unit/atomic_iteration_test.lua:165: expected:
{}
actual:
{
    [1, '1', null, null, 14],
    [2, '2', null, null, 14],
    [3, '3', null, null, 14],
    [4, '4', null, null, 14],
    [5, '5', null, null, 14],
    [6, '6', null, null, 14],
    [7, '7', null, null, 14],
    [8, '8', null, null, 14],
    [9, '9', null, null, 14],
    [10, '10', null, null, 14],
}
stack traceback:
	...orkspace/expirationd/test/unit/atomic_iteration_test.lua:165: in function 'retrying'
	...orkspace/expirationd/test/unit/atomic_iteration_test.lua:164: in function 'atomic_iteration.engine:"vinyl".index_type:"TREE".test_mvcc_vinyl_tx_conflict'
	...
	[C]: in function 'xpcall'
```

Tarantool commit that introduced the issue: https://github.com/tarantool/tarantool/commit/b9f6d385840da82f1f0a7963325fe0cff60a293e.

It looks like the test should be reworked.

oleg-jukovec:
The patch changes the logic of the test. It makes the test more determined. The patch also adds comments into the test to make it easier to understand.

Closes #105


test/unit/atomic_iteration_test.lua
```diff
@@ -134,39 +134,48 @@ end
...
-    t.skip('Skipped until tarantool/expirationd#105 is resolved')
 
     for i = 1,10 do
         cg.space:insert({i, tostring(i), nil, nil, 0})
     end
 
     local updaters = {}
     for i = 1,10 do
-        local updater = fiber.create(function()
-            fiber.name(string.format("updater of %d", i), { truncate = true })
```
Totktonada, line 147: Nit: The rule of thumb is always set a fiber name. This way it is easier to understand what is going wrong if something is going wrong.

Totktonada: Ah, I see, it is set below. Usually we do it in a first line of a fiber function.

oleg-jukovec: Thank you! I'll try to remember.

test/unit/atomic_iteration_test.lua
```diff
@@ -134,39 +134,49 @@ end
...
 
     local updaters = {}
-    for i = 1,10 do
-        local updater = fiber.create(function()
+    for i = 1,tuples_cnt do
+        local updater = fiber.new(function()
             fiber.name(string.format("updater of %d", i), { truncate = true })
-            while true do
+            box.atomic(function()
                 cg.space:update({i}, { {"+", 5, 1} })
```
olegrok, line 148: You don't need to wrap update to box.atomic. Single update operation is in transaction anyway.

test/helper.lua
```diff
@@ -201,9 +201,6 @@ t.before_suite(function()
         wal_dir    = t.datadir,
         memtx_dir  = t.datadir,
         vinyl_dir  = t.datadir,
-        -- vinyl_memory is changed to test test_mvcc_vinyl_tx_conflict.
-        -- TODO: Commented out until tarantool/expirationd#105 is resolved.
-        -- vinyl_memory = 1024,
```
olegrok, line 206: It was needed to force select operation perform read operation from disk and yield.
It's worth to see in commit message why disabling of deferred delete changes behaviour.

test/unit/atomic_iteration_test.lua
olegrok: It also looks strange because you use atomic_iteration = true... Seems you wrap transaction into transaction (it's strange that it doesn't lead to any errors).

oleg-jukovec: > You don't need to wrap update to box.atomic. Single update operation is in transaction anyway.

Thanks for the simplification. I originally had a different idea, a bit confused as a result.

oleg-jukovec: > It also looks strange because you use atomic_iteration = true... Seems you wrap transaction into transaction (it's strange that it doesn't lead to any errors).

```
-- @boolean[opt] options.atomic_iteration

False (default) to process each tuple as a single transaction and true
to process tuples from each batch in a single transaction.
```

So we have a `select tuples + delete` tx (from expirationd). It is in conflict with update transactions from fibers. Everything looks good for me.

olegrok: Yes, you are right. It's different transactions. I was a little confused.

olegrok (approved):

GRISHNOV (approved):
Hi! LGTM

PR #122, opened 2022-07-13
api: process a task on a writable space by default

### LINKED ISSUE #42: Refine master-master mode support
Initially the module leant on `box.cfg.replication_source` (`box.cfg.replication` for tarantool-1.7.6+) to don't start task processing on a replica. Here the word `replica` means an instance with at least one configured upstream.

Then something in the world changes and it becomes usual to use bidirectional replication with tarantool. Most of times it is not a real master-master: all writes are going to a single instance at given moment of the time. But it allows to skip replication reconfiguration step when switching an instance that aimed to process writes. This way decreases a time for the switching and decreases probability of mistakes during replica set reconfiguration. `box.cfg{read_only = true|false}` is often used to protect instances, where writes should not be performed.

In this confguration the expirationd module should process tasks only on the writeable instance and so checking of a `box.cfg.replication` is not sufficient. The option `force` [was added](https://github.com/tarantool/expirationd/issues/10) with this idea in the mind. It let a user control where to start task processing and where don't. However it obligates an application developer to externally stop tasks processing on an instance that goes to read-only and start it on an instance that goes to be writeable. It is [not quite obvious](https://github.com/tarantool/expirationd/issues/14).

This ideal solution would be ability to configure expirationd to obtain information about an instance state. Say, 'stop processing when `box.info.ro` becomes `true` and start it when it becomes `false`'.

I propose the following API:

```lua

expirationd.start(name, space_id, is_tuple_expired, {
    <...>
    when = <function>,  -- process tuples only when this condition is true;
                        -- this may be a user-defined function or a predefined one:
                        -- * `expirationd.has_no_upstreams` (it is default)
                        -- * `expirationd.is_writeable` (usually good for master-master setup)
    force = <boolean>,  -- process tuples despite `opts.when` value
})

--
-- Built-in functions to use in `opts.when`.
--
-- Implementations are listed here to make the idea more clear.
--

expirationd.has_no_upstreams = function()
    -- `replication_source` is the old name of `replication` parameter (prior to 1.7.6).
    -- `box.cfg{replication = {}}` may be used to withdraw from upstreams.
    local upstreams = box.cfg.replication or box.cfg.replication_source
    return upstreams == nil or (type(upstreams) == 'table' and next(upstreams) == nil)
end

expirationd.is_writeable = function()
    -- Not sure whether `box.info.ro` can be `nil` during box configuration.
    -- Consider `nil` as 'read-only' just in case.
    return box.info.ro == false
end
```

The discussion was started in PR #37 by @filonenko-mikhail, then was stale for a long time. I filed the issue to discuss and finally decide what would be most convenient way to configure the module in a multi-instance environment.

oleg-jukovec:
expirationd does not process a task by default if a replication is configured for a node. It does not matter the node is a master or a replica.
    
The patch improves the behavior. The behavior becomes different for RW and RO nodes [1] by default:
    
* A task processes on all types of spaces for a RW node.
* A task processes on temporary and local spaces for a RO node.
    
An user can still force the processing using the `force` option.
  
1. https://www.tarantool.io/en/doc/latest/reference/configuration/#confval-read_only
    
Closes #42


expirationd.lua
```diff
@@ -258,9 +258,29 @@ local function worker_loop(task)
...
-        local replication = box.cfg.replication_source or box.cfg.replication
-        if replication == nil or #replication == 0 or task.force then
+        if not box.info.ro or is_ro_writable or task.force then
+            if is_skipped then
+                -- Wait for maximum upstream lag * 2 if we became a master.
+                -- It helps to don't hit a conflict due to parallel processing
+                -- of remaining transactions from the old master and from the
+                -- new master.
+                local max_lag = 0
+                for _, r in ipairs(box.info.replication) do
```
olegrok, line 273: Seems you should also check that r is not nil.

AFAIK there could be gaps in case if user manually clears _cluster space. See https://github.com/tarantool/metrics/pull/71/files

oleg-jukovec: Oh, it is unexpected. Thank you!

`ipairs` not always works with too big gaps, so I prefer to use `table.maxn` this purpose:
```patch
index ccbd396..ccd7d12 100644
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -270,8 +270,9 @@ local function worker_loop(task)
                 -- of remaining transactions from the old master and from the
                 -- new master.
                 local max_lag = 0
-                for _, r in ipairs(box.info.replication) do
-                    if r.upstream and max_lag < r.upstream.lag then
+                for i = 1, table.maxn(box.info.replication) do
+                    local r = box.info.replication[i]
+                    if r and r.upstream and max_lag < r.upstream.lag then
                         max_lag = r.upstream.lag
                     end
                 end
```

README.md
```diff
@@ -55,13 +55,13 @@ You can:
...
 See API documentation in https://tarantool.github.io/expirationd/
 
-Note about using expirationd with replication: by default expirationd should
-process tasks only on the writeable instance, it means that expirationd *will
-not* start task processing on a replica. Here the word 'replica' means an
-instance with at least one configured upstream, it's an option
-`box.cfg.replication_source`. One can force running task on replica with option
-`force` in `start()` module function. The option force let a user control where
-to start task processing and where don't.
+Note about using expirationd with replication: by default expirationd process
```
olegrok, line 58: `process` -> `processes`

expirationd.lua
```diff
@@ -258,9 +258,30 @@ local function worker_loop(task)
...
-        if replication == nil or #replication == 0 or task.force then
+        if not box.info.ro or is_ro_writable or task.force then
+            if is_skipped then
+                -- Wait for maximum upstream lag * 2 if we became a master.
+                -- It helps to don't hit a conflict due to parallel processing
+                -- of remaining transactions from the old master and from the
+                -- new master.
+                local max_lag = 0
+                for i = 1, table.maxn(box.info.replication) do
+                    local r = box.info.replication[i]
```
olegrok, line 274: Probably it's better to call box.info once at loop start.
Because box.info() call produces really huge table and it increases pressure on GC.

expirationd.lua
```diff
@@ -623,11 +646,13 @@ end
 --     False (default) to process each tuple as a single transaction and true
 --     to process tuples from each batch in a single transaction.
 -- @boolean[opt] options.force
---     By default expirationd does not start task processing on an instance with
---     configured upstreams (see [`box.cfg.replication`][1]). Set the option
---     to `true` to enable task processing on the instance.
+--     By default expirationd process tasks for all types of spaces only on an
```
olegrok, line 649: `process` -> `processes`

expirationd.lua
oleg-jukovec: I don't think that RO -> RW switching to be too frequent. This is a rare event.

Also we always have a delay in the loop:
```lua
 314         -- Full scan iteration is complete, yield                               
 315         fiber.sleep(task.full_scan_delay) 
 ```
 
 I don't think it will cause any problems.

File: README.md (hunk 2 above)
Comment (oleg-jukovec) [reply README.md]: Thank you! Updated.

File: expirationd.lua (hunk 4 above)
Comment (oleg-jukovec) [reply expirationd.lua]: Thank you! Updated.

File: expirationd.lua (hunk 3 above)
Comment (oleg-jukovec) [reply expirationd.lua:274]: Unfortunately, we need to check `box.info.ro` here too. But iterations are not too frequent here (full index processing + delay).

File: test/integration/master_replica_test.lua [hunk 5]
```diff
@@ -0,0 +1,121 @@
...
+    local success_wait = false
+    for _ = 1, 30 do
+        local ret = cg.cluster:server('s1-replica').net_box:eval([[
+            return #box.space.customers:select({}, {limit = 10})
+        ]])
+        if ret == tuples_cnt then
+            success_wait = true
+            break
+        end
+        fiber.sleep(1)
```
DifferentialOrange, line 115: Tarantool replication is not so slow, I think it's better to set a smaller refresh timeout.

test/integration/master_replica_test.lua
```diff
@@ -0,0 +1,121 @@
...
+function g.test_replica_is_disabled(cg)
+    insert_tuples(cg)
+    local ret = cg.cluster:server('s1-master').net_box:eval([[
+        return #box.space.customers:select({}, {limit = 10})
+    ]])
+    t.assert_equals(ret, tuples_cnt)
+
+    -- wait tuples on the replica
+    local success_wait = false
+    for _ = 1, 30 do
```
DifferentialOrange, line 107: Why not use `helpers.retrying`?

test/unit/ro_test.lua
```diff
@@ -0,0 +1,131 @@
...
+    local task = expirationd.start("clean_all", space.id, helpers.is_expired_debug,
+            {full_scan_delay = 0})
+
+    helpers.retrying({}, function()
+        t.assert_equals(helpers.iteration_result, tuples)
+    end)
+
+    task:kill()
+    box.cfg({read_only = false})
+    space:drop()
```
DifferentialOrange, line 60: If assert fails, space wouldn't be dropped.

DifferentialOrange: On the other hand. it doesn't seems like it should break anything.

DifferentialOrange (approved):
Seems ok

test/integration/master_replica_test.lua
```diff
@@ -0,0 +1,121 @@
...
+
+g.after_all(function(cg)
+    if is_cartridge_helpers then
+        cg.cluster:stop()
+        fio.rmtree(cg.cluster.datadir)
+    end
+end)
+
+g.before_each(function()
+    t.skip_if(not is_cartridge_helpers, "cartridge is not installed")
```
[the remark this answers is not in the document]
oleg-jukovec: Yes, it is.

test/integration/master_replica_test.lua
oleg-jukovec: Thanks, it will be easier. Let's just use `helpers.retrying`.

test/integration/master_replica_test.lua
oleg-jukovec: It was rewritten to `helpers.retrying`.

test/integration/master_replica_test.lua
```diff
@@ -0,0 +1,121 @@
...
+        if deleted == %d then
+            break
+        end
+        require("fiber").yield()
+    end
+    task:kill()
+    return #box.space.customers:select({}, {limit = 10})
+]], tuples_cnt)
+
+function g.test_master_is_enabled(cg)
```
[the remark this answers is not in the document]
oleg-jukovec: What do you think about these new names?
```patch
-function g.test_master_is_enabled(cg)
+function g.test_expirationd_on_master_processing(cg)
     insert_tuples(cg)
     local ret = cg.cluster:server('s1-master').net_box:eval([[
         return #box.space.customers:select({}, {limit = 10})
@@ -95,7 +95,7 @@ function g.test_master_is_enabled(cg)
     t.assert_equals(ret, 0)
 end
 
-function g.test_replica_is_disabled(cg)
+function g.test_expirationd_on_replica_no_processing(cg)
```

test/unit/ro_test.lua
oleg-jukovec: Thank you, fixed.

expirationd.lua
oleg-jukovec: In pseudo-code the loop looks like:
```Lua
while true do
    if box.info.ro or other then
         get max lag from box.info.replication
    end
    full index scan with tuple processing and delays
    delay
end
```
I think an iteration will take a long time, so that it won't be a problem to check `box.info`.

test/integration/master_replica_test.lua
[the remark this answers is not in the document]
DifferentialOrange: Yeah, that's better

expirationd.lua
DifferentialOrange: I guess this comment is about calling box.info twice in the same iteration (for `.ro` and for `.replication`).

oleg-jukovec: Hmmm, I didn't think from this point of view. I got it. Thank you!
```lua
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -263,15 +263,16 @@ local function worker_loop(task)
     local is_ro_writable = space.temporary or space.is_local
     local is_skipped = false
     while true do
-        if not box.info.ro or is_ro_writable or task.force then
+        local box_info = box.info
+        if not box_info.ro or is_ro_writable or task.force then
             if is_skipped then
                 -- Wait for maximum upstream lag * 2 if we became a master.
                 -- It helps to don't hit a conflict due to parallel processing
                 -- of remaining transactions from the old master and from the
                 -- new master.
                 local max_lag = 0
-                for i = 1, table.maxn(box.info.replication) do
-                    local r = box.info.replication[i]
+                for i = 1, table.maxn(box_info.replication) do
+                    local r = box_info.replication[i]
                     if r and r.upstream and max_lag < r.upstream.lag then
                         max_lag = r.upstream.lag
                     end
```

GRISHNOV (approved):
Hi! LGTM

Issue #107, opened 2022-04-22
Offer cartridge role


Totktonada:
It'll easy setup periodical expiration in a cartridge application.

oleg-jukovec:
We had a pull request before. It's a simple role for tests:
https://github.com/tarantool/expirationd/pull/67

The role is not an independent unit. It just saves the states of expirationd's tasks. Tasks must be started from the client code manually. 

So I propose an another approach:

1. Launching tasks from a configuration file.
2. Arguments for launching are similar as for [expirationd.start](https://tarantool.github.io/expirationd/#start) + the role parameters.
3. The role parameter can be: restart or try to restore from previous state? Something else?

```yaml
expirationd:
  first_role_name:
    space_id: "space_id"
    is_expired: "is_expired_func"
    options:
       args: "arg"
       force: true
       full_scan_delay: 1
       etc
    resume: true
  second_role_name:
    space_id: "space_id_second"
    is_expired: "is_expired_func_second"
    options:
       args: "arg"
       force: true
       full_scan_delay: 1
       etc
    resume: false
```

We can load functions by name from ```_G``` table:

```lua
local function load_function(func_name)
    if from == nil or type(func_name) ~= 'string' then
        return nil
    end

    local func = rawget(_G, func_name)
    if func == nil or type(func) ~= 'function' then
        return nil
    end
    return func
end
```

The role function ```stop``` will be very simple:

```lua
local function stop()
    for _, task in tasks_list do
        log.info('stop expirationd task: %s', task)
        expirationd.kill(task)
    end
    tasks_list = {}
    return true
end
```

```validate_config``` and ```apply_config``` will be much more complicated. We must to validate very many parameters in ```validate_config```. ```apply_config``` is a good place to start/realod tasks. Something like that:

```lua
local function apply_config(conf, opts) -- luacheck: no unused args
    local conf = conf[role_name] or {}

    stop()
    if opts.is_master then
        start(conf)
    end

    return true
end
```

**I have two questions:**
1. What additional parameters are needed for the role? Do we need to try to resume tasks, or is it enough just to start it? Do we need a parameter for this?
2. Do we need additional methods for configuration from a client code, as it is done, for example, in [metrics.lua](https://github.com/tarantool/metrics/blob/82bf601d766d6d1d02986c38c4136facd240ea9e/cartridge/roles/metrics.lua#L246)?

But I propose to start with a simple implementation. And add additional parameters/functions later.

no1seman:
[...]

PR #123, opened 2022-07-18
api: continue a task from a last tuple

### LINKED ISSUE #54: Add test for reloading the module with external iteration state via Cartridge hotreload

oleg-jukovec:
After the patch a task continues processing from a last tuple at startup unless it was killed using `task:kill()` or `expirationd.kill()`. We cannot provide a safe behavior for all user cases, so it works only with default `start_key` and `iterate_with` callbacks.

The behavior supports a cartridge hot-reload.

Closes #54

A cartridge role is out of scope here and will be done for #107 . In the pull request just continue tasks + cartridge hot-reload tests for it.

### DIFF
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -25,6 +25,7 @@ end
 local stash_names = {
     cfg = '__expirationd_cfg',
     metrics_stats = '__expirationd_metrics_stats',
+    task_continue = '__expirationd_task_continue',
 }
 
 if is_hotreload_package then
@@ -41,6 +42,7 @@ local function stash_get(name)
 end
 
 local task_list = {}
+local task_continue = stash_get(stash_names.task_continue)
 local cfg = stash_get(stash_names.cfg)
 if cfg.metrics == nil then
     cfg.metrics = true
@@ -90,10 +92,9 @@ local function metrics_enable()
     -- Workaround for a cartridge role reload:
     --
     -- Metrics package does not lose observation after a cartridge reload since
-    -- 0.13.0. expirationd does not yet support the cartridge reload (it
-    -- requires saving and restarting tasks at least). So, we are acting here
-    -- as if all expirationd tasks have been killed: reset all collectors and
-    -- the callback.
+    -- 0.13.0. expirationd does not automatically restart tasks after the
+    -- cartridge reload. So, we are acting here as if all expirationd tasks
+    -- have been killed: reset all collectors and the callback.
     if metrics_stats.callback then
         metrics.unregister_callback(metrics_stats.callback)
     end
@@ -168,11 +169,11 @@ end
 -- Task fibers
 -- ------------------------------------------------------------------------- --
 
--- get all fields in primary key(composite possible) from tuple
-local function construct_key(space_id, tuple)
+-- get all fields in key(composite possible) from a tuple
+local function construct_key(space_id, index_id, tuple)
     return fun.map(
         function(x) return tuple[x.fieldno] end,
-        box.space[space_id].index[0].parts
+        box.space[space_id].index[index_id].parts
     ):totable()
 end
 
@@ -326,7 +327,9 @@ end
 -- @section Methods
 --
 local Task_methods = {
-    --- Start a task.
+    --- Start a task. It continues processing from a last tuple if the task
+    --  was previously stopped by @{task.stop} and the task has default
+    --  `start_key` and `iterate_with` functions.
     --
     -- @param  self
     --     Task instance.
@@ -377,6 +380,7 @@ local Task_methods = {
     -- @function task.restart
     restart = function (self)
         self:stop()
+        task_continue[self.name] = nil
         self:start()
     end,
 
@@ -399,6 +403,7 @@ local Task_methods = {
             end
         end
         task_list[self.name] = nil
+        task_continue[self.name] = nil
     end,
 
     --- Get a statistics about a task.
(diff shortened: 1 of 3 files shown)


test/integration/reload_test.lua
```diff
@@ -59,7 +62,90 @@ local function reload_roles(srv)
...
+    return tuple or {}
+]]
+
+function g.test_task_continue_after_reload(cg)
+    cg.cluster:server('s1-master').net_box:eval(task_sleep_on_10_eval)
+
+    reload_roles(cg.cluster:server('s1-master'))
+
+    local tuple = cg.cluster:server('s1-master').net_box:eval(task_first_tuple_eval)
+    t.assert_equals(tuple, {10})
```
DifferentialOrange, line 120: I don't get it. Shouldn't tuples from `{1}` to `{10}` expire on the first eval? (I don't have expirationd experience.)

expirationd.lua
```diff
@@ -25,6 +25,7 @@ end
 local stash_names = {
```
DifferentialOrange, line 25: So why we don't restart tasks after reload? (Again, I'm not experienced in expirationd.)

test/integration/reload_test.lua
oleg-jukovec: We set `process_expired_tuple`, so expirationd does not delete the tuples, just iterate.

expirationd.lua
oleg-jukovec: This is debatable, but it doesn't look like we should do it here:

1. If an user manually calls expirationd.start() from his cartridge role - nothing stops him from just doing it again after hot-reload.
2. From our own cartridge role we can write the logic.

At the expirationd-utility level is difficult to provide the correct behavior for all cases (for example, after hot-reload a config changes and user does not need the task anymore, or a space was deleted, or an index). For my taste, it's better to implement such logic on a higher level - cartridge-role-level.

Therefore, I provide only basic support of cartridge hot-reload here.

expirationd.lua
```diff
@@ -492,20 +497,92 @@ end
...
+    if key == nil or it == nil then
+        return nil, nil
+    end
+    return key, it
+end
 
 -- default iterate_with function
 local function default_iterate_with(task)
-    return task.index:pairs(task.start_key(), { iterator = task.iterator_type })
+    local continue_key, continue_it = create_continue_state(task)
```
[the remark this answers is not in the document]
oleg-jukovec: Shouldn't, `create_continue_state` is called once per iteration over an index.

A code inside iterator just copy a tuple pointer per tuple, it should be cheap too.

test/integration/reload_test.lua
```diff
@@ -59,7 +62,90 @@ local function reload_roles(srv)
...
+local task_sleep_on_10_eval = [[
+    for i = 1,100 do
+        box.space.customers:insert({i})
+    end
+    local expirationd = require('expirationd')
+    local tuples_cnt = 0
+    local is_expired_sleep = function()
+        tuples_cnt = tuples_cnt + 1
+        if tuples_cnt == 10 then
+            require("fiber").sleep(60)
```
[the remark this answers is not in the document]
oleg-jukovec: Thank you, done.

test/integration/reload_test.lua
```diff
@@ -59,7 +62,90 @@ local function reload_roles(srv)
...
+        end
+        return true
+    end
+    task = expirationd.start("walk_all", box.space.customers.id, is_expired_sleep,
+                             {process_expired_tuple = function() return true end,
+                              force = true})
+
+    for i = 1,100 do
```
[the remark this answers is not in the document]
oleg-jukovec: It works fine ok because we have a 60-seconds sleep inside is_expired call.

We need to wait just a very short time to iterate over 10 tuples in the memory.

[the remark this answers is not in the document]
oleg-jukovec: try:
```
--- a/test/integration/reload_test.lua
+++ b/test/integration/reload_test.lua
@@ -82,12 +82,6 @@ local task_sleep_on_10_eval = [[
                              {process_expired_tuple = function() return true end,
                               force = true})
 
-    for i = 1,100 do
-        if tuples_cnt >= 10 then
-            break
-        end
-        fiber.yield()
-    end
 ]]
 
 local task_first_tuple_eval = [[
@@ -104,13 +98,6 @@ local task_first_tuple_eval = [[
     task = expirationd.start("walk_all", box.space.customers.id, is_expired_tuple,
                              {force = true})
 
-    for i = 1,100 do
-        if tuple ~= nil then
-            break
-        end
-        fiber.yield()
-    end
-
     task:kill()
 
     return tuple or {}
@@ -118,6 +105,8 @@ local task_first_tuple_eval = [[
 
 function g.test_task_continue_after_reload(cg)
     cg.cluster:server('s1-master').net_box:eval(task_sleep_on_10_eval)
+    local tuple = cg.cluster:server('s1-master').net_box:eval(task_first_tuple_eval)
+    t.assert_equals(tuple, {10})
 
     reload_roles(cg.cluster:server('s1-master'))
 
@@ -129,6 +118,8 @@ function g.test_task_continue_after_stop_and_reload(cg)
     cg.cluster:server('s1-master').net_box:eval(task_sleep_on_10_eval .. [[
         task:stop()
     ]])
+    local tuple = cg.cluster:server('s1-master').net_box:eval(task_first_tuple_eval)
+    t.assert_equals(tuple, {10})
 
     reload_roles(cg.cluster:server('s1-master'))
 
@@ -140,6 +131,8 @@ function g.test_task_not_continue_after_kill_and_reload(cg)
     cg.cluster:server('s1-master').net_box:eval(task_sleep_on_10_eval .. [[
         task:kill()
     ]])
+    local tuple = cg.cluster:server('s1-master').net_box:eval(task_first_tuple_eval)
+    t.assert_equals(tuple, {10})
 
     reload_roles(cg.cluster:server('s1-master'))
```

test/integration/reload_test.lua
oleg-jukovec: Because it does not actually iterate over {10} tuple, it sleep on it.

test/integration/reload_test.lua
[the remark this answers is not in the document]
oleg-jukovec: It will be fragile because ```local tuple = cg.cluster:server('s1-master').net_box:eval(task_first_tuple_eval)``` actually starts a new task and it can change the state.

test/unit/continue_test.lua
```diff
@@ -0,0 +1,242 @@
+local expirationd = require("expirationd")
+local fiber = require("fiber")
+local t = require("luatest")
+local helpers = require("test.helper")
+local g = t.group('expirationd_continue')
+
+local task_name = "walk_all"
```
DifferentialOrange, line 7: You can pass this variable to evals too if you want consistency.

test/unit/continue_test.lua
```diff
@@ -0,0 +1,242 @@
...
+            end
+        end
+        return helpers.is_expired_debug(args, tuple)
+    end
+    local task = expirationd.start(task_name, space.id, is_expired,
+                                   {process_expired_tuple = function() return true end})
+    return task
+end
+
+function g.test_task_continue_after_error()
```
DifferentialOrange, line 55: So what was the expected behavior before this patch? `tuples_repeat` in the second assert?

DifferentialOrange (approved):

test/integration/reload_test.lua
DifferentialOrange: Oh, yeah, I had reread the code and now I'm see it.

test/integration/reload_test.lua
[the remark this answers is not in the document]
DifferentialOrange: I still don't get why we can't use the same approach as in unit test.

test/unit/continue_test.lua
oleg-jukovec: Yes, it is.

test/unit/continue_test.lua
oleg-jukovec: I guess you mean something like that:
```patch
--- a/test/integration/reload_test.lua
+++ b/test/integration/reload_test.lua
@@ -62,7 +62,8 @@ local function reload_roles(srv)
     t.assert_equals({ok, err}, {true, nil})
 end
 
-local task_sleep_on_10_eval = [[
+local walk_task_name = "walk_all"
+local task_sleep_on_10_eval = string.format([[
     local fiber = require("fiber")
 
@@ -78,7 +79,7 @@ local task_sleep_on_10_eval = [[
         end
         return true
     end
-    task = expirationd.start("walk_all", box.space.customers.id, is_expired_sleep,
+    task = expirationd.start("%s", box.space.customers.id, is_expired_sleep,
                              {process_expired_tuple = function() return true end,
                               force = true})
 
@@ -88,9 +89,9 @@ local task_sleep_on_10_eval = [[
         end
         fiber.yield()
     end
-]]
+]], walk_task_name)
 
-local task_first_tuple_eval = [[
+local task_first_tuple_eval = string.format([[
     local fiber = require("fiber")
 
@@ -101,7 +102,7 @@ local task_first_tuple_eval = [[
         end
         return true
     end
-    task = expirationd.start("walk_all", box.space.customers.id, is_expired_tuple,
+    task = expirationd.start("%s", box.space.customers.id, is_expired_tuple,
                              {force = true})
 
     for i = 1,100 do
@@ -114,7 +115,7 @@ local task_first_tuple_eval = [[
     task:kill()
 
     return tuple or {}
-]]
+]], walk_task_name)
```

test/integration/reload_test.lua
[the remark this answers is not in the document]
oleg-jukovec: We have nothing to wait for at the test instance of Tarantool.

At the cluster instance I can replace `fiber.yield()` by `helpers.repeat()` inside evals if you want. It does not look like a critical change. Done:

https://github.com/tarantool/expirationd/blob/cc7605aba3dfd2820716f65facb4456b29af7681/test/integration/reload_test.lua#L87-L92

test/unit/continue_test.lua
DifferentialOrange: This is a valid approach. The other one is
```lua
local eval_str = [[
    local arg = ...
    return func(arg)
]]

net_box.eval(eval_str, {arg})
```

test/integration/reload_test.lua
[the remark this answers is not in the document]
DifferentialOrange: Oh, yeah, now I get what the struggle was. I think it should be okay now.

GRISHNOV (approved):
Hi! LGTM

PR #126, opened 2022-07-26
api: wait until a space or an index is created

### LINKED ISSUE #64: Infinity increase of fiber counter

### LINKED ISSUE #68: expirationd.lua:394: attempt to index a nil value
Seems if we try to run a task for non-existent space_id we get such non-informative error

### LINKED ISSUE #116: Add conditional launch of expirationd tasks
Often, when cluster is bootstrapped from the scratch there is no any user spaces at all and developer needs to implement custom logic to start expirationd task only after space(s) to be proceesing by task will be created.

It wiil be great if this kind of machinery will be already onboard.

oleg-jukovec:
After the patch an user can to start expirationd task disregarding whether a space (and a primary/requested index) is exists or not. expirationd does not start a worker fiber until the space or index appears.

Also, after the patch expirationd stops all task fibers if a space or an index configuration error happens:

* the space or the index disappears (drop/rename);
* an unsupported type of the index;
* an invalid iterator or a start key for the index;

Closes #64
Closes #68
Closes #116

### DIFF
--- a/expirationd.lua
+++ b/expirationd.lua
@@ -168,12 +168,73 @@ end
 -- ------------------------------------------------------------------------- --
 -- Task fibers
 -- ------------------------------------------------------------------------- --
+local function check_space_and_index_exist(task)
+    local space = box.space[task.space_id]
+    if space == nil then
+        local prefix = "Space with " ..
+                       (type(task.space_id) == "string" and "name " or "id ")
+        return false, prefix .. task.space_id .. " does not exist"
+    end
+
+    local index = space.index[task.index]
+    if index == nil then
+        local prefix = "Index with " ..
+                       (type(task.index) == "string" and "name " or "id ")
+        return false, prefix .. task.index .. " does not exist"
+    end
+
+    return true
+end
+
+local function check_space_and_index(task)
+    local ok, err = check_space_and_index_exist(task)
+    if not ok then
+        return false, err
+    end
+
+    local space = box.space[task.space_id]
+    local index = space.index[task.index]
+    if index.type ~= "TREE" and index.type ~= "HASH" then
+        return false, "Not supported index type, expected TREE or HASH"
+    end
+
+    if space.engine == "memtx" and index.func ~= nil then
+        local supported = false
+        local version = rawget(_G, "_TARANTOOL"):split('-', 1)[1]
+        local major_minor_patch = version:split('.', 2)
+
+        local major = tonumber(major_minor_patch[1])
+        local minor = tonumber(major_minor_patch[2])
+        local patch = tonumber(major_minor_patch[3])
+        -- https://github.com/tarantool/expirationd/issues/101
+        -- fixed since 2.8.4 and 2.10
+        if (major > 2) or (major == 2 and minor == 8 and patch >= 4)
+          or (major == 2 and minor >= 10) then
+            supported = true
+        end
+        local force_allow = task.force_allow_functional_index or false
+        if not supported and not force_allow then
+            return false, "Functional indices are not supported for" ..
+                          " Tarantool < 2.8.4, see" ..
+                          " options.force_allow_functional_index"
+        end
+    end
+
+    local ok, err = pcall(function()
+        index:pairs(task.start_key(), {iterator = task.iterator_type})
+    end)
+    if not ok then
+        return false, err
+    end
+
+    return true, nil
+end
 
 -- get all fields in key(composite possible) from a tuple
-local function construct_key(space_id, index_id, tuple)
+local function construct_key(space_id, index, tuple)
     return fun.map(
         function(x) return tuple[x.fieldno] end,
-        box.space[space_id].index[index_id].parts
+        box.space[space_id].index[index].parts
     ):totable()
 end
 
@@ -195,7 +256,8 @@ end
 
 local function suspend(task)
     -- Return the number of tuples in the space
-    local space_len = task.index:len()
+    local index = box.space[task.space_id].index[task.index]
+    local space_len = index:len()
     if space_len > 0 then
         suspend_basic(task, space_len)
     end
@@ -323,8 +385,20 @@ local function guardian_loop(task)
     fiber.name(string.format("guardian of %q", task.name), { truncate = true })
 
     while true do
+        if check_space_and_index_exist(task) then
+            break
+        end
+        fiber.sleep(constants.check_interval)
+    end
+
+    while true do
         -- if fiber doesn't exist
         if get_fiber_id(task.worker_fiber) == 0 then
+            local ok, err = check_space_and_index(task)
+            if not ok then
+                log.info("expiration: stop task %q, reason: %s", task.name, err)
+                return
+            end
             -- create worker fiber
             task.worker_fiber = fiber.create(worker_loop, task)
 
@@ -525,19 +599,21 @@ local function default_tuple_drop(space_id, args, tuple)
 end
 
 local function create_continue_key(tuple, old_parts, task)
-    if tuple == nil or #old_parts ~= #task.index.parts then
+    local index = box.space[task.space_id].index[task.index]
+
+    if tuple == nil or #old_parts ~= #index.parts then
         return nil
     end
 
     for i, part in ipairs(old_parts) do
         for k, v in pairs(part) do
-            if task.index.parts[i][k] == nil or task.index.parts[i][k] ~= v then
+            if index.parts[i][k] == nil or index.parts[i][k] ~= v then
                 return nil
             end
         end
     end
 
-    return construct_key(task.space_id, task.index.id, tuple)
+    return construct_key(task.space_id, task.index, tuple)
 end
 
 local continue_iterators_map = {}
@@ -561,15 +637,16 @@ local function create_continue_state(task)
     local key = nil
     local it = nil
     local c = task_continue[task.name]
-    if c and c.space_id == task.space_id and c.index_id == task.index.id and c.it == task.iterator_type then
+    if c and c.space_id == task.space_id and c.index == task.index and c.it == task.iterator_type then
         key = create_continue_key(c.tuple, c.index_parts, task)
         it = continue_iterators_map[task.iterator_type]
     end
 
+    local index = box.space[task.space_id].index[task.index]
     task_continue[task.name] = {
         space_id = task.space_id,
-        index_id = task.index.id,
-        index_parts = table.deepcopy(task.index.parts),
+        index = task.index,
+        index_parts = table.deepcopy(index.parts),
         it = task.iterator_type,
         tuple = nil,
     }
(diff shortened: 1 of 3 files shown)


Totktonada:
There are two opposite cases:

* A user starts a task before creating a space/an index.
* A user makes a mistake: say, typo in the space name.

In the former case the expected behavior is to wait until the space/the index appears. In the latter case one would expect that the error will be reported.

I don't know, where to draw the line. We can add some timeout to wait until the space will appear. We can write stuck tasks into logs. We can wait in `expirationd.start()` till success (or timeout) or register the task and always report success.

I have no good vision how it should look. What do you think?

oleg-jukovec:
> There are two opposite cases:
> 
>     * A user starts a task before creating a space/an index.
> 
>     * A user makes a mistake: say, typo in the space name.
> 
> 
> In the former case the expected behavior is to wait until the space/the index appears. In the latter case one would expect that the error will be reported.
> 
> I don't know, where to draw the line. We can add some timeout to wait until the space will appear. We can write stuck tasks into logs. We can wait in `expirationd.start()` till success (or timeout) or register the task and always report success.
> 
> I have no good vision how it should look. What do you think?

Good questions, I also asked these questions to myself. I did the simplest thing here: we start the task anyway and write a message to the log.

https://github.com/tarantool/expirationd/blob/be82e3cf031d95c03b0c01ff9fc2c332d64c4edb/expirationd.lua#L1014

It looks in logs like this:

https://github.com/tarantool/expirationd/blob/be82e3cf031d95c03b0c01ff9fc2c332d64c4edb/test/unit/space_index_test.lua#L45

This logic is really bad at protecting against typo. But at the same time it is simple and obvious. It seems to me that this is enough in the case of `expirationd` because an user will notice that expirationd is not working and will start looking for a problem.

Other approaches only complicate matters, but do not solve the problem:
- An user may need to wait X*2 minutes.
- An user may do not want to see a lot of garbage in the logs.
- etc

expirationd.lua
```diff
@@ -962,8 +1004,15 @@ local function expirationd_run_task(name, space_id, is_tuple_expired, options)
...
-    -- check valid of iterator_type and start key
-    task.index:pairs( task.start_key(), { iterator = task.iterator_type })
+    local ok, err = check_space_and_index_exist(task)
+    if ok then
+        local ok, err = check_space_and_index(task)
+        if not ok then
+            error(err)
+        end
+    else
+        log.info("schedule task " .. name .. ", reason: " .. err)
```
Totktonada, line 1014: Did you mean 'postpone'? 'Schedule' sounds for me as something about completely valid/expected case, but here we should warn the user. BTW, shouldn't it be a on the warnings log level?

Totktonada:
As I understood, expirationd will write a warning once at starting of a task and then let the task wait its space/index. At least, considering other possible solutions, it looks as the simplest one. So, maybe, it is the best one. Thank you for the clarification!

expirationd.lua
oleg-jukovec: Good ideas. Now the message in the log looks like this:
`W> expiration: postpone a task clean_all, reason: Index with name non_existing_name does not exist`

test/unit/space_index_test.lua
```diff
@@ -0,0 +1,301 @@
...
+            space:insert({1, "1"})
+
+            local stderr = ""
+            helpers.retrying({}, function()
+                stderr = stderr .. capture:flush().stderr
+                t.assert_str_contains(stderr, "expiration: stop task")
+            end)
+        end)
+
+        t.assert_equals(helpers.iteration_result, {})
```
DifferentialOrange, line 199: How should one interpret this block of asserts?

test/unit/space_index_test.lua
```diff
@@ -0,0 +1,301 @@
...
+        helpers.iteration_result = {}
+
+        local capture = luatest_capture:new()
+        capture:wrap(true, function()
+            case.fun(cg.space, cg.space.index[index_id])
+
+            local stderr = ""
+            helpers.retrying({}, function()
+                stderr = stderr .. capture:flush().stderr
+                t.assert_str_contains(stderr, "expiration: stop task")
```
DifferentialOrange, line 267: So what will happen after I recreate the space or the index?

test/unit/space_index_test.lua
```diff
@@ -0,0 +1,301 @@
...
+        t.assert_equals(helpers.iteration_result, {})
+        t.assert_equals(#expirationd.tasks(), 1)
+        t.assert_not_equals(expirationd.task(task_name), nil)
+        t.assert_equals(task:statistics().restarts, 1)
+
+        task:kill()
+    end
+end
+
+function g.test_run_after_truncate(cg)
```
DifferentialOrange, line 280: Why do we test this here? What possibly could change?

DifferentialOrange (approved):
Seems nice

test/unit/space_index_test.lua
oleg-jukovec: The task doesn't process any tuple.

The pattern `helpers.iteration_result + helpers.is_expired_debug` used in many tests.

test/unit/space_index_test.lua
```diff
@@ -0,0 +1,301 @@
...
+    },
+    space_rename = {
+        fun = function(_, index)
+            index:rename("XXX")
+        end
+    },
+}
+
+for name, case in pairs(rename_cases) do
+    g["test_stop_after_" .. name .. "_if_name"] = function(cg)
```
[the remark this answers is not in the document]
oleg-jukovec: Renamed:
`if_name` -> `if_name_used`
`if_id` -> `if_id_used`

test/unit/space_index_test.lua
```diff
@@ -0,0 +1,301 @@
...
+    if box.space[cg.space.name] then
+        cg.space:drop()
+    end
+    if tmp_space then
+        tmp_space:drop()
+        tmp_space = nil
+    end
+end)
+
+function g.test_start_on_existing_space_and_index(cg)
```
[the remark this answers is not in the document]
oleg-jukovec: I don't see such simple test. Something else is being checked in other tests. Besides, it looks good here.

test/unit/space_index_test.lua
```diff
@@ -0,0 +1,301 @@
...
+g.before_each(function(cg)
+    cg.space = helpers.create_space_with_tree_index('memtx')
+    cg.space:insert({1, "1"})
+    -- kill live tasks (it can still live after failed tests)
+    for _, t in ipairs(expirationd.tasks()) do
+        expirationd.kill(t)
+    end
+end)
+
+local tmp_space = nil
```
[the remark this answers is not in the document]
oleg-jukovec: Thank you, updated.

test/unit/space_index_test.lua
```diff
@@ -0,0 +1,301 @@
+local expirationd = require("expirationd")
+local fiber = require("fiber")
+local t = require("luatest")
```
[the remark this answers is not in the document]
oleg-jukovec: Thank you, updated: single -> double.

test/unit/space_index_test.lua
oleg-jukovec: And we tests that all fine after truncate. For my taste, it is ok  here.

test/unit/space_index_test.lua
oleg-jukovec: It depends on how fast you recreate space:
1. It may stop if you were slow.
2. It may continue to work.

I've added such tests.

Idea was not to identify all wrong cases and stop, but just to stop if we can't rerun a task. If we have a space and index specified in a task options - we can rerun.

test/unit/space_index_test.lua
DifferentialOrange: The task has run a single time in
```lua
helpers.retrying({}, function()
    t.assert_equals(helpers.iteration_result, {{1, "1"}})
end)
```

How we can differ between "the task has stopped" and "timeout is not yet passed for a task to run a second time, but it will run if we wait a little bit"?

test/unit/space_index_test.lua
DifferentialOrange: But rename is drop+create, not a truncate

test/unit/space_index_test.lua
oleg-jukovec: There is no way to check the state of the task from an user side. It remains only to believe the message in the logs.

test/unit/space_index_test.lua
DifferentialOrange: Thank you, I'll get it.

> 2. It may continue to work if you were quick.

To be honest, I'm not sure that it should be covered with tests.

test/unit/space_index_test.lua
oleg-jukovec: Yes, truncate is mentioned in the comments. create/drop/truncate/rename are a space manipulations with possible side-effects in the future.

Therefore, we need to add at least a basic test for truncate. Even if we are sure that now it will not lead to problems.

test/unit/space_index_test.lua
DifferentialOrange: > There is no way to check the state of the task from an user side.

Well, that's a pity. But I still don't get why in the end of both `"test_stop_after_" .. name .. "_if_name_used"` and `"test_not_stop_after_" .. name .. "if_id_used"` we assert the same `t.assert_equals(task:statistics().restarts, 1)`

test/unit/space_index_test.lua
```diff
@@ -0,0 +1,364 @@
...
+        end)
+
+        t.assert_equals(helpers.iteration_result, {})
+        t.assert_equals(#expirationd.tasks(), 1)
+        t.assert_not_equals(expirationd.task(task_name), nil)
+        t.assert_equals(task:statistics().restarts, 1)
+
+        task:kill()
+    end
+    g["test_not_stop_after_" .. name .. "if_id_used"] = function(cg)
```
DifferentialOrange, line 209:
```suggestion
    g["test_not_stop_after_" .. name .. "_if_id_used"] = function(cg)
```

test/unit/space_index_test.lua
DifferentialOrange: Well, ok, let it be

test/unit/space_index_test.lua
oleg-jukovec: I've added:
```
t.assert_equals(task.worker_fiber:status(), "dead")
t.assert_equals(task.guardian_fiber:status(), "dead")
```
into `*_stop_*` tests. These are implementation details, but we should quickly notice if changes in the implementation break our tests (we will rename the fibers, or delete a fiber).

> Well, that's a pity. But I still don't get why in the end of both "test_stop_after_smth_if_name_used" and "test_not_stop_after_smth_if_id_used" we assert the same `t.assert_equals(task:statistics().restarts, 1)`

restarts == 0 if a task execution started, restarts == 1 if a task execution started, restarts == 2 if a task execution started again (an error happens).

test/unit/space_index_test.lua
oleg-jukovec: I think it is ok, this is a test for the expected behavior.

test/unit/space_index_test.lua
oleg-jukovec: Thank you, fixed!

test/unit/space_index_test.lua
DifferentialOrange: >These are implementation details

I think it's okay for a unit test.

DifferentialOrange (approved):

no1seman (approved):

GRISHNOV (approved):
Hi! LGTM

PR #129, opened 2022-08-10
api: rename space_id parameter to space

### LINKED ISSUE #112: Confusing space_id parameter
We can use string name of space, but parameter only says id. 
Information from doc `space_id can be numeric or string` isn't not enough in my opinion.
Maybe we should add more words about it in doc. Or add new alias method.

oleg-jukovec:
expirationd.start() has the parameter space_id. The parameter name is
confusing because the value can be a name of a space, not just id.

This change does not affect an old code:

* It changes the name of the parameter, not the key in the options
  table. It doesn't break the current code.
* An user should not use a task.space_id field directly [1]. We are
  free to rename it.

1. https://github.com/tarantool/expirationd/blob/b9bcbe0a1782ffc1225c33d1eab78ef4e60162f8/expirationd.lua#L420-L423

Closes #112

It is the time to make this correction (if we really want it) so that the name doesn't get into the Tarantool Cartridge role parameter:
https://github.com/tarantool/expirationd/blob/b9bcbe0a1782ffc1225c33d1eab78ef4e60162f8/README.md?plain=1#L167

After the 1.3.0 release, it will be more difficult to change the parameter name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,8 +77,8 @@ function is_expired(args, tuple)
   return true
 end
 
-function delete_tuple(space_id, args, tuple)
-  box.space[space_id]:delete{tuple[1]}
+function delete_tuple(space, args, tuple)
+  box.space[space]:delete{tuple[1]}
 end
 
 expirationd.start(job_name, space.id, is_expired, {
@@ -164,7 +164,7 @@ package with features:
   ```yaml
   expirationd:
     task_name1:
-      space_id: 579
+      space: 579
       is_expired: is_expired_func_name_in__G
       is_master_only: true
       options:
--- a/cartridge/roles/expirationd.lua
+++ b/cartridge/roles/expirationd.lua
@@ -106,7 +106,7 @@ local function get_task_config(task_conf)
     -- setmetatable resets __newindex write protection on a copy
     local conf = setmetatable(table.deepcopy(task_conf), {})
     local params_map = {
-        space_id = {required = true, types = {"n", "s"}},
+        space = {required = true, types = {"n", "s"}},
         is_expired = {required = true, types = {"f"}},
         is_master_only = {required = false, types = {"b"}},
         options = {required = false, types = {"t"}},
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -164,7 +164,7 @@ package with features:
   ```yaml
   expirationd:
     task_name1:
-      space_id: 579
+      space: 579
```
DifferentialOrange, line 167: So for now, it will be `space_id` options field and corresponding `space` role config field?

README.md
```diff
@@ -77,8 +77,8 @@ function is_expired(args, tuple)
   return true
```
DifferentialOrange, line 77: Isn't it a time to create additional `space` input option and add a log warning that `space_id` is deprecated?

DifferentialOrange (commented):
I think you should inspect [TDG2](https://github.com/tarantool/tdg2) code to check is it isn't rely on task.space_id in any way.

DifferentialOrange:
> After the 1.3.0 release, it will be more difficult to change the parameter name.

With current Cartridge configuration implementation, parameter migration is really painful, much painful that in-code parameter migration.

README.md
oleg-jukovec: Just `space` as param-name for `expirationd.start()` and `space` as a task param for the role.

I think we can rename until we don't do a release with the role.

README.md
oleg-jukovec: I'm sorry, but I didn't get it.

`space` looks good as a required parameter, not option.

oleg-jukovec:
> I think you should inspect [TDG2](https://github.com/tarantool/tdg2) code to ensure it isn't relies on `task.space_id` in any way.

Fortunately it doesn't use `task.space_id`.

README.md
DifferentialOrange: Oh, it looks like `space`/`space_id` is a positional argument in Lua-style task.

```Lua
expirationd.start(job_name, space.id, is_expired, opts)
```

I have thought that it is a named one.

README.md
DifferentialOrange: See https://github.com/tarantool/expirationd/pull/129#discussion_r942384194

DifferentialOrange (approved):

GRISHNOV (approved):

PR #130, opened 2022-08-19
test: fix vinyl_is_broken() check

### LINKED ISSUE #103: test: vinyl_is_broken() check is not fully correct
I'm about 838c2d16346741cd753f6a95810284ecb4c6b50a and, in particular, about the following code:

https://github.com/tarantool/expirationd/blob/838c2d16346741cd753f6a95810284ecb4c6b50a/test/helper.lua#L251-L262

See, `broken_v1_10` may be `true` on 2.*, so the function will return `true` on, say, 2.11.0.

The second condition (`broken_v2`) looks correct. However I propose to consider a bit more strict way to write such checks, see https://github.com/tarantool/crud/pull/227. In brief, if a problem was fixed in, say, 1.10.2, 2.1.3, 2.2.2, 2.3.1 (and all 2.4* and all 3.* by construction), the condition will be the following:

```lua
-- `true` if it works, `false` otherwise.
(major == 1 and minor == 10 and patch >= 2) or
(major == 2 and minor == 1 and patch >= 3) or
(major == 2 and minor == 2 and patch >= 2) or
(major == 2 and minor == 3 and patch >= 1) or
(major == 2 and minor >= 4) or
(major >= 3)
```

There are less chances to make a mistake when such strict form of condition is used.

While I'm here, it worth to share the recipe how to determine, where a particular change lands in tarantool's long-term branches:

```sh
$ (needle='vinyl: fix use of dropped space in deferred DELETE handler'; for branch in 1.7 1.9 1.10 2.1 2.2 2.3 2.4 2.5 2.6 2.7 2.8 master; do printf "${branch}\t"; commit_description=$(git log --oneline "origin/${branch}" | grep -F "${needle}"); [ -n "${commit_description}" ] && echo "$(git describe --long --always $(echo "${commit_description}" | cut -d" " -f1)) $(echo "${commit_description}" | cut -d" " -f2-)" || echo "(not found)"; done)

1.7	(not found)
1.9	(not found)
1.10	1.10.11-5-g40d8f4df1 vinyl: fix use of dropped space in deferred DELETE handler
2.1	(not found)
2.2	(not found)
2.3	(not found)
2.4	(not found)
2.5	(not found)
2.6	(not found)
2.7	(not found)
2.8	2.8.2-12-g2663a9c6c vinyl: fix use of dropped space in deferred DELETE handler
master	2.10.0-beta1-51-g0428bbcef vinyl: fix use of dropped space in deferred DELETE handler
```

So, it lands to 1.10.12, 2.8.3 and will land to future 2.10.0. All 2.11+ and 3+ will be unaffected as well, of course.

oleg-jukovec:
The function may return true for Tarantool 2.11. It is incorrect
because the bug was fixed in 2.10:

     ...
     1.10   1.10.11-5-g40d8f4df1
     ...
     2.8    2.8.2-12-g2663a9c6c
     2.10   2.10.0-beta1-51-g0428bbcef
     master 2.10.0-beta1-51-g0428bbcef

The patch fixes the problem and unifies a code that checks a support
of features.

Closes #103

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Incorrect check of the Tarantool version in the tests to determine a bug in
+  the vinyl engine that breaks the tests (#103).
+
 ## 1.3.0 - 2022-08-11
 
 This release adds a Tarantool Cartridge role for expirationd package and
--- a/test/helper.lua
+++ b/test/helper.lua
@@ -258,25 +258,33 @@ function helpers.tarantool_version()
     return major, minor, patch
 end
 
-function helpers.vinyl_is_broken()
-    -- Blocked by https://github.com/tarantool/tarantool/issues/6448
+function helpers.vinyl_is_supported()
     local major, minor, patch = helpers.tarantool_version()
 
-    -- Since Tarantool 1.10.11.
-    local broken_v1_10 = major >= 1 and (minor > 10 or minor == 10 and patch >= 11)
-
-    -- Tarantool >= 2.1.0 and < 2.8.2.
-    local broken_v2 = (major >= 2 and major < 3) and (minor >= 1 and minor < 8 or minor == 8 and patch <= 2)
-
-    return broken_v1_10 or broken_v2
+    -- The issue: https://github.com/tarantool/tarantool/issues/6448
+    --
+    -- The problem was introduced in 1.10.2 and fixed in 1.10.12, 2.8.3 and
+    -- after a 2.10 release.
+    return (major == 1 and minor <= 9) or
+        (major == 1 and minor == 10 and patch <= 1) or
+        (major == 1 and minor == 10 and patch >= 12) or
+        (major == 1 and minor >= 11) or
+        (major == 2 and minor == 8 and patch >= 3) or
+        (major == 2 and minor >= 10) or
+        (major >= 3)
 end
 
-function helpers.memtx_func_index_is_broken()
+function helpers.memtx_func_index_is_supported()
     local major, minor, patch = helpers.tarantool_version()
-    if major > 2 or (major == 2 and minor == 8 and patch >= 4) or (major == 2 and minor >= 10) then
-        return false
-    end
-    return true
+
+    -- The issue: https://github.com/tarantool/tarantool/issues/6786
+    --
+    -- Functional indexes for memtx storage engine are introduced in 2.2.1 with
+    -- a bug. The 1.10 series does not support them at all. The problem was
+    -- fixed in 2.8.4 and after a 2.10 release.
+    return (major == 2 and minor == 8 and patch >= 4) or
+        (major == 2 and minor >= 10) or
+        (major >= 3)
 end
 
 return helpers
--- a/test/unit/custom_index_test.lua
+++ b/test/unit/custom_index_test.lua
@@ -10,8 +10,9 @@ local g = t.group('custom_index', {
 })
 
 g.before_each({index_type = 'TREE'}, function(cg)
-    t.skip_if(cg.params.engine == 'vinyl' and helpers.vinyl_is_broken(),
-        'Blocked by https://github.com/tarantool/tarantool/issues/6448')
+    t.skip_if(cg.params.engine == 'vinyl' and not helpers.vinyl_is_supported(),
+        'Blocked by https://github.com/tarantool/tarantool/issues/6448 on ' ..
+		'this Tarantool version')
     g.space = helpers.create_space_with_tree_index(cg.params.engine)
 end)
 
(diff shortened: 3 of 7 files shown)


test/helper.lua
```diff
@@ -258,25 +258,27 @@ function helpers.tarantool_version()
...
-    return broken_v1_10 or broken_v2
+    -- Blocked by https://github.com/tarantool/tarantool/issues/6448
+    -- Tarantool < 1.10.11 for Tarantool v1
+    -- Tarantool <= 2.1.0 or >= 2.10 for Tarantool v2
+    return (major == 1 and minor < 10) or
+        (major == 1 and minor == 1 and patch < 11) or
+        (major == 2 and minor <= 1) or
+        (major == 2 and minor >= 10) or
+        (major >= 3)
 end
```
Totktonada, line 272: 1. I would use `<=`, `>=` consistently, it is easier to interpret as 'starting from X.Y', 'ending at X.Y' -- more intuitive IMHO.
2. `minor == 1` should be `minor == 10` I guess.
3. I'm not sure I understand the comment. Whether the affected range is between introducing `_vinyl_deferred_delete` (1.10.1-309-gf4cb5fc8a) and fixing the linked issue (1.10.11-5-g40d8f4df1, 2.8.2-12-g2663a9c6c, 2.10.0-beta1-51-g0428bbcef)? IOW, supported versions are:
    * 1.10: v <= 1.10.1 and v >= 1.10.12
    * 2.8: v >= 2.8.3
    * 2.10: all
    * 3: all

Totktonada: BTW, it does not look as blocked anymore. I would rephrase it.

test/helper.lua
```diff
@@ -258,25 +258,27 @@ function helpers.tarantool_version()
...
-    if major > 2 or (major == 2 and minor == 8 and patch >= 4) or (major == 2 and minor >= 10) then
-        return false
-    end
-    return true
+
+    -- Blocked by https://github.com/tarantool/tarantool/issues/6786
+    -- Tarantool >= 2.8.4 or Tarantool >= 2.10
+    return (major == 2 and minor == 8 and patch >= 4) or
+        (major == 2 and minor >= 10) or
+        (major >= 3)
```
Totktonada, line 281: I think a short comment is needed here. The following facts may be not obvious for a reader:

* Functional indexes for memtx storage engine are introduced in 2.2.1. The 1.10 series does not support them at all.
* Vinyl is not supported yet: https://github.com/tarantool/tarantool/issues/4492.
* The https://github.com/tarantool/tarantool/issues/6786 problem was fixed in 2.8.4 and 2.10.0. We verify the related functionality only on tarantool versions that include this fix.

test/unit/custom_index_test.lua
```diff
@@ -198,7 +198,7 @@ function g.test_tree_index_multikey(cg)
 end
 
 function g.test_memtx_tree_functional_index_broken(cg)
-    t.skip_if(_TARANTOOL < "2" or not helpers.memtx_func_index_is_broken(),
+    t.skip_if(_TARANTOOL < "2" or helpers.memtx_func_index_is_supported(),
               "Unsupported Tarantool version")
```
Totktonada, line 202: It seems `_TARANTOOL < "2"` is already part of the `memtx_func_index_is_supported()` check.

Totktonada: Or I don't understand some logic here. We just check that expirationd raises a meaningful error and don't touch the space, corrent?

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Is vinyl broken check in tests (#103).
```
Totktonada, line 16: At least, please, qoute 'is vinyl broken'. Otherwise it is hard to parse against the English grammar.

oleg-jukovec: No problems, I've rewrote the message:
```
### Fixed

- Incorrect check of the Tarantool version in the tests to determine a bug in
  the vinyl engine that breaks the tests (#103).
```

test/unit/custom_index_test.lua
oleg-jukovec: > It seems `_TARANTOOL < "2"` is already part of the `memtx_func_index_is_supported()` check.

But Tarantool does not and we don't create a functional index in a test helper. I've split the condition into a two to make it clear.

oleg-jukovec: > Or I don't understand some logic here. We just check that expirationd raises a meaningful error and don't touch the space, corrent?

Yes, it is.

test/helper.lua
oleg-jukovec: Thank you, I've added more comments.

test/helper.lua
oleg-jukovec: Thank you, it's my fault. I looked at the wrong commit.

test/unit/custom_index_test.lua
Totktonada: > I've split the condition into a two to make it clear.

Nice idea, thanks.

test/unit/custom_index_test.lua
```diff
@@ -197,11 +197,14 @@ function g.test_tree_index_multikey(cg)
...
-    t.skip_if(_TARANTOOL < "2" or not helpers.memtx_func_index_is_broken(),
-              "Unsupported Tarantool version")
-    t.skip_if(cg.params.index_type ~= 'TREE', 'Unsupported index type')
+function g.test_memtx_tree_functional_index_broken_error(cg)
     t.skip_if(cg.params.engine == 'vinyl', 'Unsupported engine')
+    t.skip_if(cg.params.index_type ~= 'TREE', 'Unsupported index type')
+    t.skip_if(cg.space.index["functional_index"] == nil,
+        "Functional indexes are not supported by Tarantool")
+    t.skip_if(helpers.memtx_func_index_is_supported(),
+        "Doesn't blocked by https://github.com/tarantool/tarantool/issues/6786")
```
Totktonada, line 206: Nit: I would say something like 'no error expected with a functional index on this tarantool version, correctness is verified by another test case'. Up to you.

Totktonada (approved):

DifferentialOrange (approved):

GRISHNOV (approved):

PR #133, opened 2022-12-26
cartridge: ability to configure the role

### LINKED ISSUE #131: Add ability to pass cfg options via cartridge role configuration
Metrics cannot be enabled via the cartridge role configuration. Now we still need to write the cartridge role code. It would be great to support the feature:

```yaml
expirationd:
  cfg:
    # expirationd.cfg() params
    metrics: true
  task_name_1:
    # the task configuration
  task_name_2:
    # the task configuration
```

oleg-jukovec:
A user can configure the role using the `cfg` entry. The configuration options are the same as for expirationd.cfg().

The `cfg` entry may be a task for backward compatibility.

Closes #131

### DIFF
--- a/cartridge/roles/expirationd.lua
+++ b/cartridge/roles/expirationd.lua
@@ -137,6 +137,33 @@ local function get_task_config(task_conf)
     return conf
 end
 
+local function get_cfg(cfg)
+    local conf = setmetatable(table.deepcopy(cfg), {})
+    local params_map = {
+        metrics = {"b"},
+    }
+
+    for k, _ in pairs(conf) do
+        if type(k) ~= "string" then
+            error(role_name .. ": config option must be a string")
+        end
+        if params_map[k] == nil then
+            error(role_name .. ": unsupported config option " .. k)
+        end
+    end
+
+    for param, types in pairs(params_map) do
+        if conf[param] ~= nil then
+            local ok, res = get_param(param, conf[param], types)
+            if not ok then
+                error(res)
+            end
+        end
+    end
+
+    return conf
+end
+
 local function init()
 
 end
@@ -153,7 +180,14 @@ local function validate_config(conf_new)
         if not ok then
             error(res)
         end
-        get_task_config(task_conf)
+        local ok, ret = pcall(get_task_config, task_conf)
+        if not ok then
+            if task_name == "cfg" then
+                get_cfg(task_conf)
+            else
+                error(ret)
+            end
+        end
     end
 
     return true
@@ -174,6 +208,15 @@ local function apply_config(conf_new, opts)
             end
         end
         table.remove(started, i)
+    end
+
+    if conf["cfg"] ~= nil then
+        local ok = pcall(get_task_config, conf["cfg"])
+        if not ok then
+            local cfg = get_cfg(conf["cfg"])
+            expirationd.cfg(cfg)
+            conf["cfg"] = nil
+        end
     end
 
     for task_name, task_conf in pairs(conf) do
--- a/test/unit/role_test.lua
+++ b/test/unit/role_test.lua
@@ -8,6 +8,10 @@ local is_cartridge_roles, _ = pcall(require, 'cartridge.roles')
 local always_true_func_name = "expirationd_test_always_true"
 local iterate_pairs_func_name = "expirationd_test_pairs"
 
+g.before_all(function()
+    g.default_cfg = { metrics = expirationd.cfg.metrics }
+end)
+
 g.before_each(function()
     t.skip_if(not is_cartridge_roles, "cartridge is not installed")
 
@@ -19,11 +23,14 @@ g.before_each(function()
         g.role.kill(t)
     end
 
+    g.is_metrics_supported = helpers.is_metrics_supported()
+
     rawset(_G, always_true_func_name, function() return true end)
     rawset(_G, iterate_pairs_func_name, function() return pairs({}) end)
 end)
 
 g.after_each(function(g)
+    expirationd.cfg(g.default_cfg)
     g.space:drop()
     g.role.stop()
     for _, t in ipairs(g.role.tasks()) do
@@ -104,6 +111,31 @@ function g.test_validate_config_empty(cg)
 end
 
 local required_test_cases = {
+    cfg_empty = {
+        ok = true,
+        cfg = { ["cfg"] = {
+        }}
+    },
+    cfg_invalid_param = {
+        ok = false,
+        err = "expirationd: unsupported config option any",
+        cfg = { ["cfg"] = {
+            any = 1,
+        }}
+    },
+    cfg_metrics = {
+        ok = true,
+        cfg = { ["cfg"] = {
+            metrics = true,
+        }}
+    },
+    cfg_metrics_invalid_value = {
+        ok = false,
+        err = "expirationd: metrics must be a boolean",
+        cfg = { ["cfg"] = {
+            metrics = 12,
+        }}
+    },
     all_ok_space_number = {
         ok = true,
         cfg = { ["task_name"] = {
@@ -124,6 +156,13 @@ local required_test_cases = {
             space = 1,
             is_expired = always_true_func_name,
             options = {}
+        }}
+    },
+    all_ok_task_cfg = {
+        ok = true,
+        cfg = { ["cfg"] = {
+            space = "space name",
+            is_expired = always_true_func_name,
         }}
     },
     not_table = {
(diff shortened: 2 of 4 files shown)


GRISHNOV (approved):
Hi! LGTM

cartridge/roles/expirationd.lua
```diff
@@ -176,6 +210,15 @@ local function apply_config(conf_new, opts)
         table.remove(started, i)
     end
 
+    if conf["cfg"] ~= nil then
+        local ok = pcall(get_task_config, conf["cfg"])
+        if not ok then
+            local cfg = get_cfg(conf["cfg"])
+            expirationd.cfg(cfg)
```
DifferentialOrange, line 217: I think it's a bit trickier than this.

After this patch, user can either change configuration with `expirationd.cfg()`, which changes `expirationd.cfg` table, or with clusterwide config, which changes clusterwide config and `expirationd.cfg` table. It may be confusing that the contents of the clusterwide config section are not guaranteed to be the same as expirationd config table. This state can easily be achieved with calling `expirationd.cfg()` after applying something with clusterwide config. Also it is possible that some value X would be set only with `expirationd.cfg()` and never set explicitly in clusterwide config (it is not possible for current expirationd cfg since for now there is only one value -- `metrics`, but it is possible for `crud.cfg` which will implement similar features [soon](https://github.com/tarantool/crud/issues/332)). How can we treat it?
- The simplest case is to ignore it. Yeah, such behavior may be dissatisfying, but in most of the real cases configuration is not changed frequently (especially with two different handles). Then we must at least add a documentation warning that clusterwide config value may not represent the current `expiration.cfg{}` state and you should check `expirationd.cfg` attributes to get real values.
- We may make the behavior to be symmetric: `expirationd.cfg{}` will update clusterwide config. As far as I understand, this could be achieved with [`patch_clusterwide`](https://www.tarantool.io/ru/doc/latest/book/cartridge/cartridge_api/modules/cartridge.twophase/#patch-clusterwide-patch). The conditions of "when we should really update it" may differ: for example, we may update it if expirationd role is enabled, or if the role is enabled and there is `expirationd` section, or if the role is enabled and there is `expirationd.cfg` subsection. This one is also tricky since it is possible that `expirationd.cfg` was called before the role was enabled (but it's more about the case when we have several fields in configuration). Also I'm not sure that patching clusterwide config from the instance is a really good design of a cartridge application.
- We may simply throw error on `expirationd.cfg` call if there is a clusterwide configuration subsection `expirationd.cfg`, asking a user to continue to use clusterwide config. But it could lead to major issues on migration from `expirationd.cfg` to clusterwide config.

I see this task to be a bit more than "allow an expirationd user to set cfg with clusterwide config" -- it is likely to become an exemplar approach on how to merge `module.cfg` with clusterwide config (at least for similar task in `crud.cfg`). Maybe we should discuss this with TeamU to hear their opinion.

DifferentialOrange: As far as I understand, there are no similar examples yet -- `metrics` has `metrics.set_export`, but this handle is expected to be called at most once and has lower priority than clusterwide config (current `module.cfg` has higher priority and to change this is to introduce a breaking change).

oleg-jukovec: > After this patch, user can either change configuration with `expirationd.cfg()`, which changes `expirationd.cfg` table, or with clusterwide config, which changes clusterwide config and `expirationd.cfg` table.

My logic here is very simple:
1. clusterwide config is a global configuration to set up for all nodes by default.
2. expirationd.cfg() is a for a user/node-specific/manual configuration.

So there is nothing confusing in the fact that the values from clusterwide config and expirationd config are different.

> We may make the behavior to be symmetric: expirationd.cfg{} will update clusterwide config. As far as I understand, this could be achieved with [patch_clusterwide](https://www.tarantool.io/ru/doc/latest/book/cartridge/cartridge_api/modules/cartridge.twophase/#patch-clusterwide-patch). The conditions of "when we should really update it" may differ: for example, we may update it if expirationd role is enabled, or if the role is enabled and there is expirationd section, or if the role is enabled and there is expirationd.cfg subsection.

Here, too, everything is simple:

We may update the config only if the update is called from the role: cartridge.service_get('expirationd').cfg({}). But this is a bad idea - the configuration needs to be updated from the top-level, not from the low-level.

> We may simply throw error on expirationd.cfg call if there is a clusterwide configuration subsection expirationd.cfg, asking a user to continue to use clusterwide config. But it could lead to major issues on migration from expirationd.cfg to clusterwide config.

Ideally, the application should not know anything about the cartridge. We can only do this from the role (and I don't see the point in this restriction).

oleg-jukovec: >  but it is possible for `crud.cfg` which will implement similar features [soon](https://github.com/tarantool/crud/issues/332)). 

For which values will this be relevant and why?

oleg-jukovec: As a result:

1. A user sets the global configuration that the cartridge accepts for the module per each instance.
2. After that a user can set a local module configuration per a specific instance.

I don't see any logical problems in this logic. Yeah, the user will need to manually re-apply the local configuration of the module after cargirdge's apply_config(), but this is quite expected.

DifferentialOrange: Most people upload configuration to set up something, but it is also common to download configuration/inspect it in UI to see how everything is configured now. Current solution does not allow to do the same for expirationd since clusterwide configuration info may not be related to real cfg at all.

DifferentialOrange: > > but it is possible for `crud.cfg` which will implement similar features [soon](https://github.com/tarantool/crud/issues/332)).
> 
> For which values will this be relevant and why?

One may set up `metrics` in `crud.cfg`, but use clusterwide configuration to set `metrics_driver`.

DifferentialOrange: But your point about configuring instances differently is a valid one. Thus, it seems that we will stick to the first approach. I think we still need to add documentation warnings about configuration priority and overriding.

oleg-jukovec: > I think we still need to add documentation warnings about configuration priority and overriding.

Added:

cartridge/roles/expirationd.lua
```diff
@@ -137,6 +137,33 @@ local function get_task_config(task_conf)
     return conf
 end
 
+local function get_cfg(cfg)
```
DifferentialOrange, line 140: So in fact "get" is "validate and fill with defaults?"

DifferentialOrange: What would happen if one uploads
```yml
expirationd:
   cfg:
```
? Will defaults override existing configuration?

test/unit/role_test.lua
```diff
@@ -487,6 +519,93 @@ function g.test_apply_config_start_tasks(cg)
...
+                is_expired = always_true_func_name,
+            },
+        }
+    }, {is_master = false})
+
+    t.assert_equals(#cg.role.tasks(), 1)
+    t.assert_not_equals(cg.role.task(task_name1), nil)
+    t.assert_equals(expirationd.cfg.metrics, metrics_default)
+
+    expirationd.cfg({metrics = metrics_default})
```
DifferentialOrange, line 541: Shouldn't it be in `after_test`/`after_each` hook?

cartridge/roles/expirationd.lua
oleg-jukovec: > What would happen if one uploads
> 
> ```yaml
> expirationd:
>    cfg:
> ```
> 
> ? Will defaults override existing configuration?

It will do nothing.

test/unit/role_test.lua
oleg-jukovec: Thank you, updated.

cartridge/roles/expirationd.lua
DifferentialOrange: > > What would happen if one uploads
> > ```yaml
> > expirationd:
> >    cfg:
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ? Will defaults override existing configuration?
> 
> It will do nothing.

I get it now. Maybe we should add a simple test:
- set non-default value,
- upload `{expiractiond = cfg = {}}`,
- assert non-default value is still here.

oleg-jukovec: Added:
```test_apply_config_cfg_empty_do_nothing```

DifferentialOrange (approved):

PR #140, opened 2023-03-19
test: archive_by_timestamp

ArtDu:

Check if we can use specific is_tuple_expired, process_expired_tuple, these basic functions are included in expiration_process. Function is_tuple_expired checks the timestamp in the time field of tuple, and if current time is greater then time field then true. Function process_expired_tuple instead of simple deletion, does the archiving by deleting the tuple from the expirated space and insert into a new space that we create in before_each.

Updated test's name:

- simple expires test -> test_archive_by_timestamp

Part of #61

### DIFF
--- a/test.lua
+++ b/test.lua
@@ -282,80 +282,18 @@ init_box()
 
 -- ========================================================================= --
 -- TAP TESTS:
--- 1. simple archive test.
--- 2. errors test,
--- 3. not expire test,
--- 4. kill zombie test
--- 5. multiple expires test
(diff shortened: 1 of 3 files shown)


oleg-jukovec (approved):
Thank you for the patch!

ArtDu:
> Thank you for the patch!

I don't have a permission to merge it. Also I'll create PR with a second commit and send you on review

GRISHNOV (approved):

PR #144, opened 2023-04-02
test: check_tuples_not_expired_by_timestamp

ArtDu:

Check that after run tuples is not expired. Let's make sure that after the expiration starts, no tuples will be archived since the timestamp has an advantage of 2 seconds.

Updated test's name:

- not expired task -> test_check_tuples_not_expired_by_timestamp

Part of #61


test/unit/expiration_process_test.lua
```diff
@@ -171,3 +171,44 @@ function g.test_broken_process_expired_tuple(cg)
...
 end
+
+function g.test_check_tuples_not_expired_by_timestamp(cg)
+    local space = cg.space
+    local space_archive = cg.space_archive
+    local task_name = cg.task_name
+
+    local total = 5
+    for i = 1, total do
+        local time = fiber.time()
```
oleg-jukovec, line 182: The change will make it a little easier to find a problem in case of a flaky test + it is a logic of the original test.

```suggestion
    local total = 5
    local time = fiber.time()
    for i = 1, total do
```

oleg-jukovec (approved):
Thank you for the patch!

GRISHNOV (approved):

oleg-jukovec:
Could you please merge commits into one?

ArtDu:
> Could you please merge commits into one?

Done

oleg-jukovec:
Thank you for the patch!

PR #145, opened 2023-04-10
test: default_tuple_drop_function

ArtDu:

Check that the default expiration function will work. After expiration, there will be no tuples in the space.
All tuples will be deleted.

Updated test's name:

- default drop function test -> test_default_tuple_drop_function

Part of #61


oleg-jukovec (approved):
Thank you for the patch!

PR #150, opened 2023-05-11
test: expirationd_update

ArtDu:

Check if the expirationd module will work correctly on time and after expirationd.update.
Also fixed error in task_stop_test, a bug that affected the update test.

Updated test's names:

- restart test -> test_expirationd_update

Part of #61

### DIFF
--- a/test/unit/update_and_kill_test.lua
+++ b/test/unit/update_and_kill_test.lua
@@ -1,0 +1,96 @@
+local fiber = require("fiber")
+local expirationd = require("expirationd")
+local t = require("luatest")
+
+local helpers = require("test.helper")
+
+local g = t.group('update_and_kill', {
+    {index_type = 'TREE', engine = 'vinyl'},
+    {index_type = 'TREE', engine = 'memtx'},
+    {index_type = 'HASH', engine = 'memtx'},
+})
+
+g.before_each({index_type = 'TREE'}, function(cg)
+    t.skip_if(cg.params.engine == 'vinyl' and not helpers.vinyl_is_supported(),
+        'Blocked by https://github.com/tarantool/tarantool/issues/6448')
+    cg.first_space = helpers.create_space_with_tree_index(cg.params.engine)
+    local second_space = helpers.create_space('second_space', cg.params.engine)
+    second_space:create_index('primary')
+    cg.second_space = second_space
+end)
+
+g.before_each({index_type = 'HASH'}, function(cg)
+    cg.first_space = helpers.create_space_with_hash_index(cg.params.engine)
+    local second_space = helpers.create_space('second_space', cg.params.engine)
+    second_space:create_index('primary', {type = 'HASH'})
+    cg.second_space = second_space
+end)
+
+g.after_each(function(cg)
+    for _, task_name in pairs(expirationd.tasks()) do
+        expirationd.kill(task_name)
+    end
+    cg.first_space:drop()
+    cg.second_space:drop()
+end)
+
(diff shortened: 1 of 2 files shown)


ArtDu:
@oleg-jukovec  I rebased master and now it's green and you can check it

test/unit/update_and_kill_test.lua
```diff
@@ -0,0 +1,97 @@
...
+
+    for _, task_name in pairs(second_expd_link.tasks()) do
+        second_expd_link.kill(task_name)
+    end
+
+    -- Back old link in require. It's necessary to avoid problem of double update call.
+    -- The problem that we can't use old link properly after double update.
+    -- Old link wouldn't see new tasks which were started by new link. That could happen without this line.
+    -- Expirationd module changes only one previous link that stores in package.loaded.
+    package.loaded["expirationd"] = first_expd_link
```
oleg-jukovec, line 96: Could you give an example why this is needed?

test/unit/update_and_kill_test.lua
```diff
@@ -0,0 +1,97 @@
...
+
+        t.assert_equals(space:count(), total)
+        helpers.retrying({}, function()
+            t.assert_equals(space:count(), 0)
+        end)
+    end
+
+    for _, task_name in pairs(second_expd_link.tasks()) do
+        second_expd_link.kill(task_name)
+    end
```
oleg-jukovec, line 90: Let's move it to `after_each`:

```
    for _, task_name in pairs(expirationd.tasks()) do
        expirationd.kill(task_name)
    end
```

This will help us get better after false-assertions.

test/unit/update_and_kill_test.lua
```diff
@@ -0,0 +1,97 @@
...
+    cg.first_space = helpers.create_space_with_hash_index(cg.params.engine)
+    local second_space = helpers.create_space('second_space', cg.params.engine)
+    second_space:create_index('primary', {type = 'HASH'})
+    cg.second_space = second_space
+end)
+
+g.after_each(function(cg)
+    if cg.task ~= nil then
+        cg.task:kill()
+    end
```
oleg-jukovec, line 31: There is no cg.task in the tests, but it could be a list with all tasks `cg.tasks`.

test/unit/update_and_kill_test.lua
ArtDu: I tried to describe why it's needed in comments above. Double update happens because we run this test three times due to this:
```lua
t.group('update_and_kill', {
    {index_type = 'TREE', engine = 'vinyl'},
    {index_type = 'TREE', engine = 'memtx'},
    {index_type = 'HASH', engine = 'memtx'},
})
```

I don't want to move it in `after_each` because we don't need need this line for further kill test

test/unit/update_and_kill_test.lua
ArtDu: Moved

test/unit/update_and_kill_test.lua
oleg-jukovec: Let's do it then in `after_test`:

https://github.com/tarantool/tarantool/blob/2b1c8713b54f94b0fd3c6e106426b583c06f64ad/test/app-luatest/fiber_slice_test.lua#L20-L22

This will help to return to the original `expirationd` even if the test fails.

ArtDu: Done

oleg-jukovec (approved):
Thank you for the patch!

GRISHNOV (approved):

PR #154, opened 2023-08-18
cartridge: ability to use persistent functions

### LINKED ISSUE #153: Not able to use persistent functions
I want to be able to use functions created via box.func.create in expirationd cartridge role configuration. E.g.:

```yaml
expirationd:
  cfg:
    metrics: true
  task_name1:
    space: 579
    is_expired: box.func.my_custom_func
```

where function box.func.my_custom_func has been created via [box.func.create](https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_schema/func_create/). 

This is not working now because function created via box.func.create are not exported to _G which breaks expirationd configuration (https://github.com/tarantool/expirationd/blob/master/cartridge/roles/expirationd.lua#L10) and leads to errors like this: ```expirationd: is_expired must be a function name in _G```.

bitgorbovsky:
A user can configure the role with persistent functions as callback for task.

Closes #153

### DIFF
--- a/cartridge/roles/expirationd.lua
+++ b/cartridge/roles/expirationd.lua
@@ -2,15 +2,40 @@ local expirationd = require("expirationd")
 local role_name = "expirationd"
 local started = require("cartridge.vars").new(role_name)
 
+local function dumb_fn() end
+
 local function load_function(func_name)
     if func_name == nil or type(func_name) ~= 'string' then
         return nil
     end
 
     local func = rawget(_G, func_name)
-    if func == nil or type(func) ~= 'function' then
+    if func == nil then
+        if type(box.cfg) == 'function' then
+            -- After restart `validate_config` sometimes is run before box.cfg
+            -- call and it leads to error like `Please call box.cfg first` and
+            -- at this moment we are not able to check real availability of
+            -- functions in box.func. If we return nil, we fail configration
+            -- but it is not an error actually, because we cannot do any
+            -- checks.  Thus we decided to return dumb_fn to avoid
+            -- misconfiguration at this stage and in hope that `apply_config`
+            -- will do real check.
+            -- P.S. Of course it is a bad solution, but...
+            return dumb_fn
+        end
+        if not box.schema.func.exists(func_name) then
+            return nil
+        end
+
+        return function(...)
+            return box.func[func_name]:call({...})
+        end
+    end
+
+    if type(func) ~= 'function' then
         return nil
     end
+
     return func
 end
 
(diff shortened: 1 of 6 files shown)


cartridge/roles/expirationd.lua
```diff
@@ -2,15 +2,31 @@ local expirationd = require("expirationd")
...
 local function load_function(func_name)
     if func_name == nil or type(func_name) ~= 'string' then
         return nil
     end
 
     local func = rawget(_G, func_name)
-    if func == nil or type(func) ~= 'function' then
+    if func == nil then
+        if type(box.cfg) == 'function' then
+            return dumb_fn
```
oleg-jukovec, line 15: It seems to me that it is better to return `nil` in this case and signal about a problem. It is better to get an error at the time of checking the config than at runtime.

Or the configuration has not yet been completed at the `validate_config` step?

oleg-jukovec (changes requested):
Thank you for the feature! Let's make a couple more updates:

1. Please, add a record with the feature into the [CHANGELOG.md](https://github.com/tarantool/expirationd/blob/94c56388f7b69c70fdeabbf96dd286dd21c99f96/CHANGELOG.md?plain=1#L11).

2. Let's update [the error message](https://github.com/tarantool/expirationd/blob/master/cartridge/roles/expirationd.lua#L22) too:
`a function name in _G` -> `a function name in _G or in box.func`

3. I would like to see a unit-test with a function from `box.func`.

Something like that, but with a function name from `box.func`:
https://github.com/tarantool/expirationd/blob/94c56388f7b69c70fdeabbf96dd286dd21c99f96/test/unit/role_test.lua#L595-L608

4. I would like to see the order `a function name from _G` > `a function name from box.func` in a unit test to ensure that the function from _G will apply. It could be a valid function from `_G` + an invalid function from `box.func` with the same name + check that all works fine.

cartridge/roles/expirationd.lua
bitgorbovsky: Yes, here is a problem with restarting nodes of cartridge application.
After restart `validate_config` sometimes is run before `box.cfg` call and it leads to error like `Please call box.cfg first` and at this moment we are not able to check real availability of functions in box.func. If we return nil, we fail configration but it is not an error actually, because we cannot do any checks. Thus we decided to return dumb_fn to avoid misconfiguration at this stage and in hope that apply_config will do real check.

oleg-jukovec: Thank you for the explanation.

This is not very good, because in `apply_config` we need to have a validated (and valid) configuration. But it seems that we could not to do anything here:

https://www.tarantool.io/en/doc/latest/book/cartridge/cartridge_dev/#role-s-life-cycle-and-the-order-of-function-execution
`The validate_config() function, first, before the automatic box.cfg() call (database initialization), then – upon every configuration update.`

oleg-jukovec (approved):
Thank you for the patch!

DerekBum (approved):

PR #170, opened 2024-12-12
Fixed launch on ro instances with is_master_only parameter true

muxa-xoma:


oleg-jukovec (approved):
Thank you for the patch!

It would be nice to add an entry into CHANGELOG.md and rewrite the PR message like other commits in the repository:
```
roles: fix launch on ro instance with is_master_only = true
```
too.

PR #171, opened 2024-12-12
roles: fix launch on ro instance with is_master_only = true

muxa-xoma:


oleg-jukovec (approved):
Thank you for the patch! You could just re-push a commit into your branch instead of creating a new PR.

